PR #2, opened 2020-10-03
Feature/setup

### LINKED ISSUE #1: Setup Repo

charkour:
This looks like a lot, but I did not write any code. I just installed a bunch of NPM packages, moved files around, and formatted them with the auto-formatter. 

There is a bit going on here, so feel free to ask questions and I would love to explain what is happening. I've used this technology for many years, so it is completely normal to feel confused at the beginning. It'll make sense after some time. In terms of the code styling rules put into place, those are based on my preferences and industry norms, so feel free to suggest updates and then we can modify them to meet our team's needs. 

I'll update the readme at some point to explain all that was installed and configured, but that will be done another day. I tested this with VSCode so I recommend opening the `./vscode/*.code-workspace` file in VSCode for easy use. It'll give you recommended extensions. @Syobnaf, I know you like GEdit, so there might be extensions like ESLint or Prettier you could look into for that editor. If those don't exist, that should be fine because the command that is run whenever you commit a file will auto-format the files.

closes #1


bryantgeorge (approved):
Everything here looks really good, I ran the app (after `npm start`) and it still works (after running `npm install`). Despite usually using GEdit, I think I'll just use VSCode for this project because the extensions seem very convenient, thanks for setting those up. I tested it out a bit by modifying `index.html` in VSCode in a bad style, and when I saved, it fixed my change to follow normal style. Quite nice!

Reviewed, looks good.

kvlinden:
Fine by me - I haven't used prettier or husky, but they sound good.

PR #6, opened 2020-10-06
Develop to Production

charkour:
Merging current develop into master since there have been some milestones achieved. 

Also, this is so we can test the continuous deployment GitHub action. 

So does this count as version 0.0.1 of the app?


Jdellis818 (approved):
LGTM

bryantgeorge (approved):
Looks good

PR #7, opened 2020-10-07
add automated testing framework

### LINKED ISSUE #5: Setup Automated Testing
Through GitHub actions?

https://dev.to/dyarleniber/setting-up-a-ci-cd-workflow-on-github-actions-for-a-react-app-with-github-pages-and-codecov-4hnp might be useful here.

### LINKED ISSUE #9: Fix bug in deploy.yml
I think the     env:
      working-directory: ./client-course-schedulizer
needs to be set in `deploy.yml`. Looking at the logs, it can't find `package.json` but it is looking in the wrong directory.

bryantgeorge:
See the check that passed. I based it on how Charlie setup the Turtle Tracker testing. I would personally recommend testing on pushes to any branch, but I saw you didn't do it on the Turtle tracker, so maybe we don't want to. Let me know your thoughts on this.

Closes #5 
Closes #9

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,23 @@
+name: Install and Test
+
+on:
+  pull_request:
+    branches:
+      - develop
+      - production
+
+jobs:
+  test-client-app:
+    runs-on: ubuntu-latest
+    env:
+      working-directory: ./client-course-schedulizer
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install npm dependencies
+        run: |
+          npm ci
+        working-directory: ${{env.working-directory}}
+      - name: Run tests
+        run: |
+          npm test
+        working-directory: ${{env.working-directory}}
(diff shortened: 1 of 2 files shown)


charkour:
Looks good! I just set it up in the past to only test when there are PRs targeting `production` or `develop` branches since those are typically the only ones used on a small project like this. If we want, we could make it run on any PR that is made (or even on every commit). Since we are an open-source project, we have unlimited testing minutes through GitHub so we can really do whatever. What do you think @Syobnaf?

@Jonri2, @kvlinden, any thoughts on when/how often to do testing?

Thanks again for setting all of these up. It really speeds up development work and saves us from headaches by not letting us merge breaking changes.

bryantgeorge:
If we have no limit, then I personally would just have it test on any push because I find testing on all commits to be helpful at work, but it's not a huge deal to me

bryantgeorge:
@charkour just so you know I added testing on push, as well as fixing (hopefully) the issue with the `deployment.yml` (did this because I didn't want to create a separate PR for such a small change)

Jdellis818 (approved):
LGTM

PR #12, opened 2020-10-14
Add data structure interfaces

### LINKED ISSUE #10: Create Interfaces for Data Structures

bryantgeorge:
Branch adds implementations of the datatypes from our Google spreadsheet as typescript interfaces.

Nothing to test, but verify that the datatypes make sense and aren't missing anything important.

Should more values be optional, or will we fill in data with defaults?

Closes #10

### DIFF
--- a/client-course-schedulizer/public/dataInterfaces.tsx
+++ b/client-course-schedulizer/public/dataInterfaces.tsx
@@ -70,3 +70,7 @@ interface Section {
   studentHours?: number;
   year: number;
 }
+
+interface Schedule {
+  sections: Section[];
+}
--- a/client-course-schedulizer/public/dataInterfaces.tsx
+++ b/client-course-schedulizer/public/dataInterfaces.tsx
@@ -11,6 +11,17 @@ enum Half {
   Second = "second",
 }
 
+enum Day {
+  Friday = "F",
+  Monday = "M",
+  // TODO: Are weekends ever used? (Can't really hurt to have them)
+  Saturday = "S",
+  Sunday = "Su",
+  Thursday = "Th",
+  Tuesday = "T",
+  Wednesday = "W",
+}
+
 interface Location {
   building: string;
   roomCapacity: number;
@@ -18,8 +29,8 @@ interface Location {
 }
 
 interface Meeting {
-  // Like "MWThF" (all days on which the given Meeting time and room is applicable)
-  days: string;
+  // All days on which the given Meeting time and room is applicable
+  days: Day[];
   // In minutes (usually 50)
   length: number;
   location: Location;
@@ -52,6 +63,7 @@ interface Section {
   letter: string;
   localMax: number;
   // Multiple Meetings possible if time/room differs on different days
+  // Asynchronous classes should have an empty array of meeting times
   meetings: Meeting[];
   semester: Semester;
   // Overrides Course value
--- a/client-course-schedulizer/public/dataInterfaces.tsx
+++ b/client-course-schedulizer/public/dataInterfaces.tsx
@@ -22,6 +22,7 @@ interface Meeting {
   days: string;
   // In minutes (usually 50)
   length: number;
+  location: Location;
   // Like "8:00AM" or "12:30PM"
   startTime: string;
 }
--- a/client-course-schedulizer/public/dataInterfaces.tsx
+++ b/client-course-schedulizer/public/dataInterfaces.tsx
@@ -1,0 +1,59 @@
+enum Semester {
+  Fall = "FA",
+  Interim = "IN", // TODO: Remove?
+  Spring = "SP",
+  Summer = "SU", // TODO: Is this a thing?
+}
+
+enum Half {
+  First = "first",
+  Full = "full",
+  Second = "second",
+}
+
+interface Location {
+  building: string;
+  roomCapacity: number;
+  roomNumber: string;
+}
+
+interface Meeting {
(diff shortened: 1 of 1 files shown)


bryantgeorge:
Whoa, Netlify gives a deployment preview, that's really cool (look in checks and click on `Details` for the Netlify preview check)

client-course-schedulizer/public/dataInterfaces.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface Course {
+  facultyHours: number;
+  name: string;
+  number: string;
+  prefix: string[];
+  studentHours: number;
+}
+
+interface Section {
```
charkour, line 54: A way to signify async classes. No meeting time?

client-course-schedulizer/public/dataInterfaces.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface Location {
+  building: string;
+  roomCapacity: number;
+  roomNumber: string;
+}
+
+interface Meeting {
+  // Like "MWThF" (all days on which the given Meeting time and room is applicable)
+  days: string;
```
charkour: I suggest we make days be an array of Days. It will make things easier to parse.

charkour (changes requested):

client-course-schedulizer/public/dataInterfaces.tsx
bryantgeorge: 038a906 should address this, good idea

client-course-schedulizer/public/dataInterfaces.tsx
bryantgeorge: Put a comment in 038a906. I think we just want to have those courses have an empty array of Meeting objects, and then if the registrar needs it to be printed as a field in the spreadsheet we can just extrapolate based on courses which have no meeting times. (I wonder how we will visualize these though. They will show up in the faculty loads tab, but I don't think they would appear anywhere else, so it might be difficult for a user to go back and edit them/know that they exist. Maybe we need a place for them to be shown in the UI?)

client-course-schedulizer/public/dataInterfaces.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+enum Day {
+  Friday = "F",
+  Monday = "M",
+  // TODO: Are weekends ever used? (Can't really hurt to have them)
+  Saturday = "S",
+  Sunday = "Su",
+  Thursday = "Th",
+  Tuesday = "T",
```
charkour, line 21: I like the inclusion of the weekends just in case. Should we make Sunday and Thursday have a unique identifier? like N and R? And then we have a method to convert it to "Su" and "Th" when the schedule is output and returned to the registrar? Would it make our parsing easier?

Since it is an enum, parsing might not matter since we will be able to access the values via `Day.Thursday`

client-course-schedulizer/public/dataInterfaces.tsx
```diff
@@ -0,0 +1,76 @@
+enum Semester {
+  Fall = "FA",
+  Interim = "IN", // TODO: Remove?
+  Spring = "SP",
+  Summer = "SU", // TODO: Is this a thing?
```
charkour, line 5: I think it would be good to have these for now. However, by looking at the [new schedule](https://calvin.edu/dA/66abbcbe5f/Conceptual%20draft%20of%202021-2022%20academic%20calendar.pdf?language_id=1), does our current data model account for the four intensives, half-semester classes, and full-sesmeter classes? We might need to add some fields somewhere. Any suggestions?

client-course-schedulizer/public/dataInterfaces.tsx
charkour: Yeah, good idea. Maybe they will show up at the top of the schedule with no time on the UI. This is in reference to classes without a meeting time. Not sure why the conversation didn't link.

charkour (changes requested):
I added a couple comments to clarify a few things. Looks good overall!

client-course-schedulizer/public/dataInterfaces.tsx
bryantgeorge: I think parsing of our values wont matter much because they are stored as an array of these day strings (not as one string with multiple days). Parsing will come in for the CSVs we receive and we wont really be able to control the way they format Thursday (and Sunday), so I think the parsing "problem" will exist no matter how we store Thursday and Sunday.

client-course-schedulizer/public/dataInterfaces.tsx
bryantgeorge: That's a good graphic, I didn't know the plan was so... complex. I wonder if we would just have to change the `Half` enum to have `Full`, `Session 1`, `Session 2`, `Intensive A`, `Intensive B`, `Intensive C`, and `Intensive D` options? Is this something we should change before the January scheduling?

client-course-schedulizer/public/dataInterfaces.tsx
charkour: Yeah, that makes sense. Thanks!

client-course-schedulizer/public/dataInterfaces.tsx
charkour: Hmm, I think we can leave it as-is for now, and circle back around if we need to update it.

charkour (approved):
LGTM. Thanks for the great work.

PR #14, opened 2020-10-21
Add some basic UI layout

Jdellis818:


bryantgeorge:
Just a note that the Netlify preview seems to have failed for this, but it could just be something wrong with Netlify/the setup there.

client-course-schedulizer/src/components/Searchbar/Searchbar.tsx
```diff
@@ -0,0 +1,32 @@
+import { TextField } from "@material-ui/core";
+import { Autocomplete } from "@material-ui/lab";
+import React from "react";
+import "./Searchbar.scss";
+
+const styles = {
+  searchBar: {
+    width: 300,
+  },
+};
```
charkour: We should decide if we are going to use .scss or JSS to style our components and then make it consistent. What would you like to do?

client-course-schedulizer/src/components/AddSectionButton/AddSectionButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const handleClick = (event: MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  const open = Boolean(anchorEl);
+  const id = open ? "simple-popover" : undefined;
```
charkour: Either we should move this logic to be it's own Hook for future popover use or we can this package to manage the popover state. https://www.npmjs.com/package/material-ui-popup-state

client-course-schedulizer/src/components/ColorSelector/ColorSelector.tsx
```diff
@@ -0,0 +1,23 @@
...
+  };
+
+  return (
+    <div>
+      <InputLabel id="label">Color By</InputLabel>
+      <Select id="select" labelId="label" onChange={handleColorChange} value={colorValue}>
+        <MenuItem value={10}>Level</MenuItem>
+        <MenuItem value={20}>Room</MenuItem>
+        <MenuItem value={30}>Instructor</MenuItem>
+        <MenuItem value={40}>Prefix</MenuItem>
```
charkour: Perhaps it would be easier to maintain if we have an enumeration defined in TS that had the values `Level`, `Room`, `Instructor`, `Prefix`. Then we could use those values for these menu items and in the useState hook at the top.

client-course-schedulizer/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const Header = () => {
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+
+  const handleClick = (event: MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
```
charkour: Repeated logic, see other comment.

charkour (changes requested):
Looks really good! I like to menu bar with the Calvin Logo and the popover used to add a course. I removed the bootstrap styling since we are using Material UI.

A couple things to change: 

- Should we use the JSS styles that material UI provides or should we use SCSS and `className` to style?
- There is some repeated logic with the Popover things. Either we can turn it into a custom hook or use a package to manage that boilerplate code. 
- Make an enum for the menu item values.
- Update the file structure. I would put most of the components in the `reuseables` folder. Or if you don't want to do that, make a folder to put the related components in. 

Looks good. Thanks for the work!

client-course-schedulizer/src/components/Searchbar/Searchbar.tsx
Jdellis818: I definitely want to use scss, but I couldn't get it to work in this case. I don't know why

client-course-schedulizer/src/components/Header/Header.scss
```diff
@@ -0,0 +1,8 @@
+.toolbar {
+  display: flex;
+  justify-content: space-between;
+}
+
+.calvin-logo {
+  width: 60px;
+}
```
charkour: When possible, it would be good to make the names as generic as possible, like `org-logo` instead of `calvin-logo` and have the width be dynamic instead of static. Not going to require this change but good to keep in mind.

client-course-schedulizer/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { ScheduleToolbar } from "../Toolbar/ScheduleToolbar";
+import "./Tabs.scss";
+
+interface TabPanelProps {
+  index: number;
+  value: number;
+}
+
+const TabPanel = (props: PropsWithChildren<TabPanelProps>) => {
+  const { children, value, index, ...other } = props;
```
charkour, line 12: Good use of the ES rest operator.

client-course-schedulizer/src/components/Toolbar/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -0,0 +1,86 @@
...
+      </Grid>
+      <Grid container spacing={4}>
+        <Grid item xs>
+          <FormControl style={styles.row}>
+            <FormLabel component="legend">Days</FormLabel>
+            <FormControlLabel control={<Checkbox />} label="M" />
+            <FormControlLabel control={<Checkbox />} label="T" />
+            <FormControlLabel control={<Checkbox />} label="W" />
+            <FormControlLabel control={<Checkbox />} label="R" />
+            <FormControlLabel control={<Checkbox />} label="F" />
```
charkour: For something like this, you could make an array `days: string[] = ['m', 't', 'w', etc]` and then call `days.map((day): () => JSX { return <FormControlLabel control={<Checkbox />} label={day} /> });`. It makes it easier to change which days are shown or not and saves some lines. Same for the repeated code below.

client-course-schedulizer/src/components/Toolbar/ColorSelector/ColorSelector.tsx
```diff
@@ -0,0 +1,30 @@
+import { InputLabel, MenuItem, Select } from "@material-ui/core";
+import React, { ChangeEvent, useState } from "react";
+import "./ColorSelector.scss";
+
+enum ColorBy {
+  Level,
+  Room,
+  Instructor,
+  Prefix,
+}
```
charkour, line 10: Great use of Enum

charkour (approved):
Looks good. I'll approve it and you can merge it in. I like the use of the modern JavaScript syntax. I made a couple comments about how we should strive to use dynamic sizing rather than static and gave some tips on how to repeat less code. 

Awesome job. It's looking good!

client-course-schedulizer/src/components/Header/Header.scss
Jdellis818: In this case, I think static may be better. It's pretty weird to have the logo shrink when the window width decreases

client-course-schedulizer/src/components/Toolbar/Searchbar/Searchbar.scss
```diff
@@ -0,0 +1,3 @@
+.searchbar {
+  width: 300px;
+}
```
Jdellis818, line 3: I tried to do this multiple times and it didn't work. I'm not really sure what's different now, but I'm glad it's working.

client-course-schedulizer/src/components/Header/Header.scss
charkour: Ah good call.

PR #18, opened 2020-10-28
Parse csv into Interface

### LINKED ISSUE #13: Convert CSV into Interface Types

bryantgeorge:
This branch adds the ability for a CSV to be parsed and turned into a schedule (one course per row). To test, try `npm start` then bring up the browser's JS console, and expand the Sections object which is logged to the console. You should see a Section for each row in the new `modelSchedule.csv` file (ensure the object makes sense for what the `modelSchedule.csv` contains).

The branch also adds a `modelSchedule.csv` file for testing purposes (also in our google drive folder as a sheet), and this branch moves and slightly modifies the `dataInterfaces.ts` file to `src`.

One final note is that there are two things this branch doesn't do:
1. It doesn't reuse instructors or courses when they are identical (maybe this isn't a problem?).
2. It doesn't allow for multiple meetings in one section (we'll probably have to have some discussion on how to parse this and what that would look like in the csv).

Closes #13


client-course-schedulizer/src/index.tsx
```diff
@@ -15,3 +16,4 @@ ReactDOM.render(
 // unregister() to register() below. Note this comes with some pitfalls.
 // Learn more about service workers: https://bit.ly/CRA-PWA
 serviceWorker.unregister();
+read.readCSV();
```
charkour, line 19: This is a great way to test the readCSV, but generally, we will not edit `index.tsx`. Somewhere in the `<App />` component, there will be something that calls `read.readCSV()` once a file is uploaded. We can remove this when we add that functionality.

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
+import papa from "papaparse";
+import * as di from "./dataInterfaces";
```
charkour: Just so the `src/` folder doesn't get too cluttered, perhaps you can move the interfaces to `utilities/interfaces` or to another folder of your choosing.

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+const firstReg = RegExp("[Ff]irst");
+const secondReg = RegExp("[Ss]econd");
+const fullReg = RegExp("[Ff]ull");
+const sunReg = RegExp("[Ss][Uu]|[Nn]");
+const monReg = RegExp("[Mm]");
+const tuesReg = RegExp("[Tt](?![Hh])");
+const wedReg = RegExp("[Ww]");
+const thursReg = RegExp("[Tt][Hh]|[Rr]");
+const friReg = RegExp("[Ff]");
+const satReg = RegExp("[Ss](?![Uu])");
```
charkour: I think the time regex is great and the others are helpful, maybe we should convert all of the text to uppercase or lowercase to ease the complexity? Since you've written it out already, I think this is probably fine to leave it as is.

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+  let year: number;
+  let term: di.Term;
+  let half: di.Half;
+  let days: di.Day[];
+  let globalMax: number;
+  let localMax: number;
+  let anticipatedSize: number;
+  let instructors: di.Instructor[];
+  let comments: string;
+  // eslint-disable-next-line prefer-const
```
charkour: This can be removed since you are using const.

client-course-schedulizer/src/readCSV.ts
charkour: Also, I would put `readCSV.ts` somewhere in `utilities/`

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+    sections: [],
+  };
+
+  // Get data and fields from the CSV
+  const { data, meta } = objects;
+  const { fields } = meta;
+
+  // From the CSV fields, take the ones which we recognize
+  // eslint-disable-next-line prefer-const
+  let usableFields: string[] = [];
```
charkour: this is a weird JS thing, but if you don't reassign an array, but just push values to it, then it is a `const` array.

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+  let field;
+  let startTimeIdx = -1;
+  let startTimeStrIdx = -1;
+  let instructorIdx = -1;
+  let instructorsIdx = -1;
+  let prefixIdx = -1;
+  let prefixesIdx = -1;
```
charkour: I believe you are able to do chained assignment, like `let startTimeIdx = startTimeStrIdx = -1;` but will all of the values. Could save some lines (but not a major issue; more of a note).

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+    globalMax = 30;
+    localMax = 30;
+    anticipatedSize = 30;
+    instructors = [];
+    comments = "";
+
+    // Iterate through the fields of the CSV, and parse their values for this object
+    for (let j = 0; j < usableFields.length; j += 1) {
+      field = usableFields[j];
+      if (field in object) {
```
charkour: Since `usableFields` is taken from the parsed object, wouldn't `field` always be in `object`?

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+                  numHourPart -= 12;
+                  hourPart = String(numHourPart);
+                }
+                // Assume PM when 12:XX or military time
+                ampm = "PM";
+              }
+
+              // If hour is 0, assume military time of 12 AM
+              if (numHourPart === 0) {
+                hourPart = "12";
```
charkour: If a case is more than just two lines (assignment and break), I would suggest refactoring this code to become a helpful function that the case calls to help with the readability.

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+              term = di.Term.Fall;
+            } else if (summerReg.test(value)) {
+              term = di.Term.Summer;
+            } else if (springReg.test(value)) {
+              term = di.Term.Spring;
+            } else if (interimReg.test(value)) {
+              term = di.Term.Interim;
+            } else {
+              // eslint-disable-next-line no-console
+              console.log(`Term of "${value}" is unreadable, defaulting to Fall`);
```
charkour: Repeated structure like this within code usually signifies that it can be refactored into a function that is more flexible. For example, you could in the case call a function that tests the `value` and when it is true it assigns `term = di.Term[term]` where term is calculated by the function based on when regular expression resulted to true. This provides more flexible code. For example, if we were to get another term called Winter, then all you would have to do is add a new RegularExpression for Winter and update the Interface and the function would take care of the rest. (This is kinda a dumb example because the terms are likely to be only these four, but this thinking can be applied to the "half", and "days" below. But since it is working, maybe we just leave it as is?)

charkour (changes requested):
This looks great! It is awesome that you got this working. I did leave some comments for you to look at and address but they are all related to readability/maintainability/flexibility. The code you've written works well but it is just difficult to read or maintain in some places. I think @Jonri2 will have a few comments as well.

The comments aren't meant to discourage but rather they are to help and provide another way of approaching the problems. Writing maintainable code will help us if we need to make changes in the future. This work is really great and I'm excited to get this integrated into the rest of our app! Great use of the cool CSV parsing package. I haven't heard of that before. Also, thanks for telling me how to test the PR. Most people (like myself) forget to do that when making PRs.

client-course-schedulizer/src/readCSV.ts
charkour: There are a few other cases where it would be good to abstract the code to its own function.

charkour:
Ah, in response to your two comments, I think it is okay that it doesn't check that two classes could be identical. Since we want to allow them to update the CSV with whatever schedule and then we would display the conflicts on the tool. And sounds good, we can talk tomorrow about parsing the multiple meetings. Again, this is great work. I feel like we are ahead of schedule, right?

client-course-schedulizer/src/readCSV.ts
Jdellis818: I also think it would probably be cleaner here to do `usableFields.forEach((field) => {... loop logic ...})`

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+  let roomCapacity: number;
+  let year: number;
+  let term: di.Term;
+  let half: di.Half;
+  let days: di.Day[];
+  let globalMax: number;
+  let localMax: number;
+  let anticipatedSize: number;
+  let instructors: di.Instructor[];
+  let comments: string;
```
Jdellis818: I think it may be better to encapsulate these variables in a `SpreadsheetSection` interface. That way all you need to initialize in this function is one `SpreadsheetSection` object rather than 20 or so variables.

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+        }
+      }
+    }
+  }
+
+  // If both "startTime" and "startTimeStr" are present, ignore "startTime"
+  // This is because Pruim's data has a timestamp in "startTime" (not our timezone)
+  if (startTimeIdx !== -1 && startTimeStrIdx !== -1) {
+    usableFields.splice(startTimeIdx, 1);
+  }
```
Jdellis818: Use `fields.forEach()`. It's also probably better to make a list of possible duplicates and not to keep track of all the indexes for splicing. That way, we can adjust the possible duplicates if the spreadsheet requirements change simply by adding to the duplicates array. In practice, I think it could look something like this
```
const usableFields: string[] = [];
const duplicates = [
    ["startTime", "startTimeStr"],
    ["prefix", "prefixes"],
    ["instructor", "instructors"],
];
if (fields) {
  fields.forEach((field) => {
      if (validFields.includes(field)) {
        usableFields.push(field);
      }
  });
}

duplicates.forEach((duplicate) => {
  if (duplicate[0] in usableFields && duplicate[1] in usableFields) {
    usableFields.splice(usableFields.indexOf(duplicate[0]), 1);
  }
});
```

client-course-schedulizer/src/readCSV.ts
```diff
@@ -0,0 +1,428 @@
...
+    roomCapacity = 30;
+    year = new Date().getFullYear();
+    term = di.Term.Fall;
+    half = di.Half.Full;
+    days = [di.Day.Monday, di.Day.Wednesday, di.Day.Friday];
+    globalMax = 30;
+    localMax = 30;
+    anticipatedSize = 30;
+    instructors = [];
+    comments = "";
```
Jdellis818: Set `SpreadsheetSection` object here.

Jdellis818 (dismissed):
I think this is such a good start and it works great! I just added a few comments that should help with maintainability.

client-course-schedulizer/src/readCSV.ts
charkour: This is cool since it makes our code more flexible rather than fragile. I think you don't even need to have `if (fields)` if you are doing `fields.forEach()` since if `fields` is [], then nothing is run.

Jonri2: I just realized that the first loop can be better written as
````
const usableFields = validFields.filter((field) => {
    return validFields.includes(field);
});
```

Jdellis818: I just realized that the first loop can be better written as
```
usableFields = fields.filter((field) => {
     return validFields.includes(field);
 });
```

client-course-schedulizer/src/index.tsx
bryantgeorge: To be addressed in #21

client-course-schedulizer/src/readCSV.ts
bryantgeorge: In hindsight, this probably would have been a good idea, but I agree that since it's already all there, it might just be best to leave it as is for now.

client-course-schedulizer/src/readCSV.ts
bryantgeorge: 85a0bbd

client-course-schedulizer/src/readCSV.ts
bryantgeorge: `eslint` is setup to get mad at multi-assignment, so I think I'll leave it as is for now. (For reference the rule is `no-multi-assign`)

client-course-schedulizer/src/readCSV.ts
bryantgeorge: e28ffc3

client-course-schedulizer/src/readCSV.ts
bryantgeorge: Great idea: e28ffc3

client-course-schedulizer/src/readCSV.ts
bryantgeorge: e28ffc3

client-course-schedulizer/src/readCSV.ts
bryantgeorge: e28ffc3

client-course-schedulizer/src/readCSV.ts
bryantgeorge: Thanks a ton for the code here, I wasn't quite sure how best to write this part, but your code is way more compact (one thing I had to change was that `duplicate[0] in usableFields` needed to be `usableFields.includes(duplicate[0])` because `in` only looks at the indexes of the array for some reason).

a413c2b

client-course-schedulizer/src/readCSV.ts
bryantgeorge: a413c2b

client-course-schedulizer/src/readCSV.ts
bryantgeorge: Tried doing something like this for a while, but had some difficulty with it, will just leave it as is for now.

client-course-schedulizer/src/readCSV.ts
bryantgeorge: fa95adb and 4c83ce3

bryantgeorge:
Thanks for the review everyone, should be ready for merging now barring any other comments

charkour (approved):
This looks great. Thanks for the good work! I made one commit to use more modern JS syntax which allows for "spreading". It effectively takes the inner properties of an object and copies them to a new object.

PR #19, opened 2020-10-28
Add Import Button

### LINKED ISSUE #15: Implement Input CSV Button

Jdellis818:
This branch adds functionality to import a CSV file. At the moment, it console logs the file contents. Once Bryant's parser is merged, it will call that function rather than console logging.

Closes #15


client-course-schedulizer/src/components/Header/Header/Header.scss
```diff
@@ -6,3 +6,11 @@
 .org-logo {
   width: 60px;
 }
+
+.import-label {
+  color: black !important;
```
charkour: As a reminder, only use `!important` when needed.

client-course-schedulizer/src/components/Header/ImportButton/ImportButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  }, [file]);
+
+  return (
+    <InputLabel className="import-label" htmlFor="import-button">
+      <Input
+        className="hidden"
+        id="import-button"
+        onChange={(e: ChangeEvent<HTMLInputElement>) => {
+          e.target.files && setFile(e.target.files[0]);
+        }}
```
charkour: This is just my preference, but for OnChange handlers, it is better to define a function outside of the component rather than doing it inline. So you could do 
```
  const onChange = (e: ChangeEvent<HTMLInputElement>) => {
    e.target.files && setFile(e.target.files[0]);
  };
```

charkour (changes requested):
Looks great. Nicely done! It'll be great to integrate with Bryant's system.

I have a comment about readability you can address. Also, I can upload any file, not just a .CSV. Should you address that here or will we do that in another PR? Nice work.

Also, thanks for telling me how to test it. @Syobnaf, might have some comments as well on this PR?

client-course-schedulizer/src/components/Header/Header/Header.scss
Jdellis818: Yes, I don't think there's any way to apply these styles to a material ui component with overrising the already defined styles with `!important`

charkour (approved):
LGTM

PR #22, opened 2020-10-30
Add Form Logic to the Add/Update Section Popover

### LINKED ISSUE #16: Implement Add Section Popup Logic
https://react-hook-form.com/ and https://www.npmjs.com/package/yup could be useful. A tutorial using react-hook-form with Material UI: https://react-hook-form.com/get-started

Jdellis818:
This PR adds react-hook-form to handle the state of the AddSectionPopup and adds the data fields which were missing when the popup was originally added. The object created from submitting the data is logged in the console. That data object still needs to be typed, which will be easy once the interfaces get merged into develop. Validation also still needs to be done, but will be done in a different PR.

Closes #16


client-course-schedulizer/src/components/Toolbar/AddSectionButton/AddSectionButton.tsx
```diff
@@ -22,6 +22,7 @@ export const AddSectionButton = () => {
           horizontal: "left",
           vertical: "bottom",
         }}
+        PaperProps={{ style: { maxWidth: "50%", minWidth: "500px" } }}
```
charkour, line 25: Possible to remove inline styles here? Maybe not because it is passed as props.

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.scss
```diff
@@ -1,3 +1,7 @@
 .popover-container {
   padding: 10px;
```
charkour: Switch this to 1em rather than px.

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -1,88 +1,144 @@
...
-  Radio,
-  RadioGroup,
-  TextField,
-} from "@material-ui/core";
-import React from "react";
+import { yupResolver } from "@hookform/resolvers/yup";
+import { Button, Grid, Typography } from "@material-ui/core";
+import React, { ChangeEvent, useState } from "react";
+import { useForm } from "react-hook-form";
+import * as Yup from "yup";
```
charkour: As a note, it would be better to use destructured importing rather than importing all. It only will import used code which decreases the overall bundle size which makes the app load faster. So instead of `import *`, you could do `import {array, object}`. I think that works with Yup. Don't spend too much time trying to fix it, we can always go back later.

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -1,88 +1,144 @@
...
 import "./AddSectionPopover.scss";
 
 export const AddSectionPopover = () => {
-  const days = ["M", "T", "W", "R", "F"];
-  const terms = ["Fall", "Interim", "Spring", "May"];
+  const days = ["M", "T", "W", "Th", "F"];
+  const terms = ["FA", "SP", "SU"];
+  const semesterLengths = ["Full", "Half", "Intensive"];
+  const halfSemester = ["First", "Second"];
+  const intensiveSemester = ["A", "B", "C", "D"];
```
charkour: Once the types it merged in, we might want to see if we can iterate over the enums or types or interfaces. But this works for now.

client-course-schedulizer/src/components/reuseables/GridItem/GridItemTextField/GridItemTextField.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <Grid item xs>
+      <TextField inputRef={register} label={label} name={label.toLowerCase()} {...textFieldProps} />
+    </Grid>
+  );
+};
+
+GridItemTextField.defaultProps = {
+  textFieldProps: {},
+};
```
charkour, line 24: This is some good, clean code.

charkour: The only thing, would it be better to make an interface for the type, or is it fine to leave it as an object? I used to always make interfaces, but maybe that doesn't matter when the type will never be reused. What do you think?

client-course-schedulizer/src/components/reuseables/GridItem/GridItemCheckboxGroup/GridItemCheckboxGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+import "./GridItemCheckboxGroup.scss";
+
+export const GridItemCheckboxGroup = ({
+  label,
+  name,
+  options,
+  register,
+}: {
+  label: string;
+  name?: string | undefined;
```
charkour: Out of curiosity, since name is optional, doesn't it allow for undefined as a type, or does it need to be specified? This isn't an issue, I'm just wondering.

charkour (changes requested):
Wow, I am very impressed with how you were able to integrate and customize the forms and components. There were just a couple comments to address in addition to a couple questions that I didn't know the answer to.

The only issue I am seeing is how you needed to pass `register` to nearly every component. I'm not good enough with React to know of a way to pass a prop to a parent wrapper and have it apply to all children. This might be a fun project to figure out if we have time (but don't do that now haha) and I think it relates to the idea of [subcomponents](https://dev.to/shayanypn/buckle-with-react-sub-component-10ll) but I am not sure. Leaving this here as a note to us in the future.

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
Jdellis818: Whoops yeah for sure. I think I copied that import statement from a tutorial

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
Jdellis818: Yeah we may want to wait to merge this PR until the enums/interfaces are merged in

client-course-schedulizer/src/components/reuseables/GridItem/GridItemTextField/GridItemTextField.tsx
Jdellis818: I'll make an interface, it's probably better practice

client-course-schedulizer/src/components/Toolbar/AddSectionButton/AddSectionButton.tsx
Jdellis818: I couldn't figure out how to have a class apply through passing props. Granted I didn't try for too long

client-course-schedulizer/src/components/reuseables/GridItem/GridItemTextField/GridItemTextField.tsx
Jdellis818: Actually, I realized that there's no need for an interface since StandardTextFieldProps is already imported. I'll just set the default value to undefined rather than an empty object

client-course-schedulizer/src/components/reuseables/GridItem/GridItemCheckboxGroup/GridItemCheckboxGroup.tsx
Jdellis818: Yep, you're right

Jdellis818:
This may be a bug just for my machine, but if I click on intensive or half, the submit button goes off my screen and I can't scroll down to click it

Jdellis818:
> The only issue I am seeing is how you needed to pass `register` to nearly every component. I'm not good enough with React to know of a way to pass a prop to a parent wrapper and have it apply to all children. This might be a fun project to figure out if we have time (but don't do that now haha) and I think it relates to the idea of [subcomponents](https://dev.to/shayanypn/buckle-with-react-sub-component-10ll) but I am not sure. Leaving this here as a note to us in the future.

Would this be an appropriate place to pass register via context?

charkour:
@Jonri2, context could work but this way currently is fine with passing registers.

client-course-schedulizer/src/components/Toolbar/AddSectionButton/AddSectionButton.tsx
charkour: No worries. Just leave it.

client-course-schedulizer/src/components/reuseables/GridItem/GridItemRadioGroup/GridItemRadioGroup.tsx
```diff
@@ -0,0 +1,54 @@
...
+  onChange,
+  options,
+  register,
+}: {
+  control: ReturnType<typeof useForm>["control"];
+  defaultValue: string;
+  label: string;
+  lowercase?: boolean;
+  name?: string;
+  onChange?: (e: ChangeEvent<HTMLInputElement>) => void | undefined;
```
charkour: `| undefined` is redundant here (I think).

client-course-schedulizer/src/utilities/helpers/readCSV.ts
```diff
@@ -359,12 +364,21 @@ const convertToInterface = function convertToInterface(objects: papa.ParseResult
       // TODO: Allow for multiple meetings
       ...sss,
     };
-    schedule.sections.push(section);
+
+    const existingCourse: di.Course[] = schedule.courses.filter((c) => {
```
charkour, line 369: Perhaps add a comment as to what this is doing. A little hard to read.

client-course-schedulizer/src/utilities/interfaces/dataInterfaces.ts
```diff
@@ -48,29 +70,29 @@ export interface Course {
...
   // Overrides Course value
   studentHours?: number;
   term: Term;
-  year: number;
+  year: number | string;
 }
 
 export interface Schedule {
-  sections: Section[];
+  courses: Course[];
```
charkour, line 97: So a Schedule is a series of Courses where each course can have multiple Sections, right?

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -1,88 +1,252 @@
...
+  SemesterLength,
+  SemesterLengthOption,
+  Term,
+} from "../../../utilities/interfaces/dataInterfaces";
+import { GridItemCheckboxGroup } from "../GridItem/GridItemCheckboxGroup";
+import { GridItemRadioGroup } from "../GridItem/GridItemRadioGroup";
+import { GridItemTextField } from "../GridItem/GridItemTextField";
 import "./AddSectionPopover.scss";
 
+interface SectionInput {
```
charkour, line 21: This is similar to the type `di.Section` right? Maybe that can be used? Maybe not and that's okay.

charkour (approved):
This looks good! Just a couple comments to look at before you merge it. They are just questions or things to consider and nothing you have to change. Thanks.

Oh, and there are some conflicts with the packages. Shouldn't be hard to fix.

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
Jdellis818: I tried to do it with `di.Section` but I think that would mean the data from the form would need to be mapped using the yup resolver, which seemed a bit difficult. Maybe something to consider for the future though

client-course-schedulizer/src/utilities/interfaces/dataInterfaces.ts
Jdellis818: Yep 👍

PR #23, opened 2020-10-31
Add Faculty Calendar

### LINKED ISSUE #17: Look into Packages (UI, Calendar, Text Validation)

charkour:
This PR adds multiple configured and customized FullCalendars that represent the faculty schedule. FullCalendar is particular about how imports are sorted which conflicted with our rules. I struggled to find a working alternative so I got fed up and removed most of the import sorting rules. I don't think removing those rules is a huge issue, but I am open to recommendations. 

I ran into one CSS bug I couldn't solve and that was to determine the width of each individual calendar. I found that it is currently 600px wide but I want to make them dynamic rather than static. 

closes #17 

If you have issues running into ESLint, let me know and I'll tell you how to troubleshoot it. Make sure to run npm i in the client folder.

### DIFF
--- a/client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
+++ b/client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
@@ -10,5 +10,12 @@
 }
 
 .calendar {
-  width: 600px; // This is static, and I don't like it, but I don't know how to make it better
+  width: 25vw;
+}
+
+.calendar-title {
+  // background-color: #f2f2f2;
+  background-color: lightblue;
+  z-index: 500;
+  padding: 1em;
 }
--- a/client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
+++ b/client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
@@ -1,31 +1,26 @@
-.calendars-wrapper {
+@import "../../../styles/constants.scss";
+
+.schedule-wrapper {
   overflow-x: auto;
 }
 
-.calendars-adjacent {
+.adjacent {
   width: fit-content;
   display: flex;
   flex-direction: row;
   flex-wrap: nowrap;
 }
 
-.calendar {
+.calendar-width {
   width: 25vw;
 }
 
 .calendar-title {
-  // TODO: MAKE CSSS VARIABLES for bg color and padding/margin dependencies
-  background-color: #f2f2f2;
-  z-index: 500;
-  padding: 1em;
+  padding: $spacing;
 }
 
-.calendar-title-row {
-  margin-left: -2em;
-  padding-left: 2em;
-  background-color: #f2f2f2;
-}
-
-.top-thing {
-  top: 10vh;
+.schedule-header-row {
+  margin-left: -2 * $spacing;
+  padding-left: 2 * $spacing;
+  background-color: $bg-header;
 }
--- a/client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
+++ b/client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
@@ -14,8 +14,18 @@
 }
 
 .calendar-title {
-  // background-color: #f2f2f2;
-  background-color: lightblue;
+  // TODO: MAKE CSSS VARIABLES for bg color and padding/margin dependencies
+  background-color: #f2f2f2;
   z-index: 500;
   padding: 1em;
+}
+
+.calendar-title-row {
+  margin-left: -2em;
+  padding-left: 2em;
+  background-color: #f2f2f2;
+}
+
+.top-thing {
+  top: 10vh;
 }
--- a/client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
+++ b/client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
@@ -1,0 +1,14 @@
+.calendars-wrapper {
+  overflow-x: auto;
+}
+
+.calendars-adjacent {
+  width: fit-content;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: nowrap;
+}
+
+.calendar {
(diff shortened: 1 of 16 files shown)


client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
```diff
@@ -0,0 +1,14 @@
...
+.calendars-adjacent {
+  width: fit-content;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: nowrap;
+}
+
+.calendar {
+  width: 600px; // This is static, and I don't like it, but I don't know how to make it better
+}
```
Jdellis818: Would it be possible to make this smaller to fit another professor on the screen?

client-course-schedulizer/src/components/Toolbar/ScheduleToolbar/ScheduleToolbar.scss
```diff
@@ -4,6 +4,16 @@
   justify-content: space-between;
 }
 
+.schedule-toolbar {
+  position: sticky;
```
Jdellis818, line 10: I like the sticky toolbar!

client-course-schedulizer/src/components/reuseables/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,47 @@
...
+import moment from "moment";
+
+// Plugins
+import interactionPlugin from "@fullcalendar/interaction"; // needed for dayClick
+import timeGridPlugin from "@fullcalendar/timegrid";
+
+import React from "react";
+import "./Calendar.scss";
+
+export const initialDate = "2000-01-02";
```
Jdellis818, line 12: It seems kind of weird to need to set a date when we're not dealing with specific dates. Is there a way to make these events just with a time and not a date? It's fine if there's not.

client-course-schedulizer/src/components/reuseables/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,47 @@
...
+        events={events}
+        headerToolbar={false}
+        height="auto"
+        initialDate={initialDate}
+        initialView="timeGridWeek"
+        nowIndicator={false}
+        plugins={[interactionPlugin, timeGridPlugin]}
+        selectable
+        weekends={false}
+      />
```
Jdellis818: Could you add `slotMinTime="07:00:00"` and `slotMaxTime="22:00:00"` to limit the hours?

client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
Jdellis818: Also, is it possible to make the day labels "M", "T", "W", "Th", and "F"?

Jdellis818 (changes requested):
I really like the package. It seems like it'll likely work well for what we need. I'm a little concerned about the styling limitations, but hopefully we can overcome that. Most of my comments don't require any changes, I was just thinking about little things that aren't too important but could make it better. The only change I really want to see is limiting the hours to 7:00AM-10:00PM. Great work!

client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
Jdellis818: `25vw` seems to work for me other than the overlapping day headings. Sometimes it's weird right after I change the scss file, but if I reload the page it's adjusts nicely.

client-course-schedulizer/src/components/reuseables/Calendar/Calendar.tsx
Jdellis818: Also, is it possible to make the prof headers sticky?

client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const FacultySchedule = () => {
+  return (
+    <>
+      <ScheduleToolbar />
+      <div className="calendars-wrapper">
+        <div className="calendars-adjacent">
+          {professors.map((prof, index) => {
+            return (
+              <div key={prof} className={`calendar ${index !== 0 ? "hide-axis" : ""}`}>
```
Jdellis818: This is good, but it makes the first professor's calendar a bit smaller than the others

client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.scss
charkour: Good call, I didn't try reloading the page. It works great! Thanks.

client-course-schedulizer/src/components/reuseables/Calendar/Calendar.tsx
charkour: Unfortunately no. We can just provide functions that hide this nuance when dealing with events so we don't have to worry about it going forward.

client-course-schedulizer/src/components/reuseables/Calendar/Calendar.tsx
charkour: Yes. It is way too difficult but I managed to do just that.

client-course-schedulizer/src/components/Tabs/FacultySchedule/FacultySchedule.tsx
charkour: Yeah, I'm not sure the best way to fix this. I'll leave a TODO and recan revisit this later.

charkour:
Getting the Schedule Header to scroll horizontally and still remain sticky to the top of the screen proved to be quite challenging. This may have been one of the most technical challenges I've completed. I think it turned out well. This is kinda bad practice, but I left some todos in this PR that aren't major, but could be addressed in the future when we are polishing the app. Please let me know what you think!

Jdellis818 (approved):
The spacing of the names are a little weird, but that's definitely an issue we can tackle later on. Great work! LGTM

Issue #20, opened 2020-10-29
Export as CSV


bryantgeorge:
Maybe have this branch export as well? (If it's not too complex)

Jdellis818:
@Syobnaf I'd like to note that I made a slight change to the interface that may affect this process. Rather than having the section be the parent element, I switched it to be the Course so that we don't have to keep duplicating course data for every new section that is created. I updated the import function accordingly, but it may require a slight change in exporting as well. #22 contains the changes.

bryantgeorge:
Awesome, thanks for the heads up!

PR #24, opened 2020-11-03
Integrate Import with CSV to Schedule and Schedule to CSV

### LINKED ISSUE #21: Integrate CSV Importing and Parsing
Get the imported CSV to be parsed and converted into a schedule object

### LINKED ISSUE #26: Create Schedule Context

bryantgeorge:
This branch does a few things:
1. Feeds the CSV gathered from the import button into the code in `readCSV.ts`.
2. Updates and simplifies some code in `readCSV.ts` (particularly with regard to the recent changes to the interfaces).
3. Adds an initial version of schedule to CSV conversion code in `writeCSV.ts`.
4. Updates `modelSchedule.csv` to reflect the new interfaces as well as removing some accidentally committed stuff.

For #20 this leaves a couple tasks:
1. Have the export button actually export the generated CSV text as a file. (Where will the schedule object will be stored, once gathered from the imported CSV?)
2. Improve on the schedule to CSV conversion code such that commas wont mess up the CSV. (Necessary or not? If necessary, still maybe not worth it right now due to rarity?)

To Test: click the import button, and import `modelSchedule.csv`. The console should log the schedule object followed by the CSV text which results from converting that schedule object (using the code in `writeCSV.ts`). Try using something like [Diff Checker](https://www.diffchecker.com/) to compare the logged CSV text with the contents of `modelSchedule.csv`, and notice that they are identical.

Closes #21

### DIFF
--- a/client-course-schedulizer/public/modelSchedule.csv
+++ b/client-course-schedulizer/public/modelSchedule.csv
@@ -1,7 +1,7 @@
-name,prefixes,number,section,studentHours,facultyHours,startTime,startTimeStr,duration,location,roomCapacity,year,term,half,days,globalMax,localMax,anticipatedSize,instructors,comments
-Logic and Complexity,MATH;CS;PHIL,312,A,4,4.5,blarg,12:30 PM,50,NH 232,30,2020,FA,Full,MWThF,30,20,27,Paul Erdos,This class counts for both CS and Math credit.
-Logic and Complexity,MATH;CS;PHIL,312,B,4,4.5,blarg,1:30 PM,50,NH 232,30,2020,FA,Full,MWThF,30,20,27,Paul Erdos,This class counts for both CS and Math credit.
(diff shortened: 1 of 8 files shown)


Jdellis818 (approved):
LGTM. Though I think in a future PR we should try to make the read CSV function more flexible to allow for headings with different names, such as the ones on the math-schedule spreadsheet. I made a small change to the code that handles duplicates so it can handle more than just 2. This way we can easily add a bunch of valid fields representing the same value.

charkour (approved):
LGTM! Thanks for the great work. I like the use of the ternary operators and the ?? when coalescing the potentially undefined fields.

Two notes:

* This doesn't work well with the `1920schedule.csv` but I think it's in a different format than what is actually used. It is an aggregate of the whole year. 
* We'll need to make the exporting only be done when we click the export button.

charkour:
I added a global ScheduleContext that allows for accessing and manipulating the Schedule from any component. I can talk through my code during the meeting today.

PR #27, opened 2020-11-05
Update Import CSV

Jdellis818:
WIP - I'm working on making the import function a bit more versatile. Now to add a new field, you just need to add it to the `ValidFields` interface and add a callback (hopefully already defined in the `caseFunctions.ts` file) to the callbacks object. Let me know if you guys think this is better than it was. I set it up so it can now import `math-schedule.csv` mostly successfully. I just still haven't set it up to parse `duration` or `semesterLength`.

### DIFF
--- a/client-course-schedulizer/src/utilities/helpers/readCSV.ts
+++ b/client-course-schedulizer/src/utilities/helpers/readCSV.ts
@@ -110,9 +110,9 @@ export const csvStringToSchedule = (csvString: string): di.Schedule => {
       });
 
       // Check if the meeting is empty, and should be removed
-      // if (firstMeeting.days === [] || firstMeeting.duration === 0) {
-      //   section.meetings = [];
-      // }
+      if (firstMeeting.days === [] || firstMeeting.duration === 0) {
+        section.meetings = [];
+      }
 
       // Check if there is already a course in the schedule with the same prefix and number
       const existingCourse: di.Course[] = schedule.courses.filter((c) => {
--- a/client-course-schedulizer/src/utilities/helpers/readCSV.ts
+++ b/client-course-schedulizer/src/utilities/helpers/readCSV.ts
@@ -3,7 +3,7 @@ import * as di from "../interfaces/dataInterfaces";
 import * as cf from "./caseFunctions";
 
 interface ValidFields {
-  [key: string]: Function;
+  [key: string]: (value: string, data: cf.CaseCallbackParams) => void;
 }
 
 const callbacks: ValidFields = {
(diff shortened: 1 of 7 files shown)


client-course-schedulizer/src/utilities/helpers/readCSV.ts
```diff
@@ -294,126 +97,42 @@ export const csvStringToSchedule = (csvString: string): di.Schedule => {
...
-      return (
-        c.prefixes.every((p) => {
-          return course.prefixes.includes(p);
-        }) && c.number === course.number
-      );
-    });
+      // Check if the meeting is empty, and should be removed
+      // if (firstMeeting.days === [] || firstMeeting.duration === 0) {
+      //   section.meetings = [];
+      // }
```
charkour: Does this need to be removed or un-commented?

client-course-schedulizer/src/utilities/helpers/readCSV.ts
```diff
@@ -294,126 +97,42 @@ export const csvStringToSchedule = (csvString: string): di.Schedule => {
...
-      schedule.courses[schedule.courses.indexOf(existingCourse[0])].sections.push(section);
-    }
-    // Otherwise, add the new course to the schedule
-    else {
-      course.sections.push(section);
-      schedule.courses.push(course);
+      // If there is, add the new section to that course
+      if (existingCourse.length > 0) {
+        const existingCourseIndex = schedule.courses.indexOf(existingCourse[0]);
+        schedule.courses[existingCourseIndex].sections.push(section);
```
charkour, line 138: Thanks for making this more readable.

charkour (changes requested):
I think builds on Bryant's previous work well, which is what teamwork is all about! With that said, it is a little different than the original implementation, but I like the updates to it. What I appreciate about this update is to add a new "Valid Field", you just have to add a new function and then mapping in the `callbacks` object. In addition, I can see how this built off of the switch statement with "cases" which is a nice way to break down this problem.

Before I approve it, there is one comment to address and then also should there be unit tests for this in this PR or will that come in a future PR? If in a future PR, please make a ticket.

Nice work! I am very impressed by the updates to make reading CSV easy to update and maintain. @Syobnaf might have additional comments.

bryantgeorge:
Upon thinking about this, I have a couple thoughts:
1. I think overall this change is worth it, due to the gains in maintainability.
2. The one area where I am slightly concerned is that this code no longer has a concept of preference between similar headers, so for example if the headers `startTime` and `startTimeStr` both exist in a CSV, this code will just utilize the data in whichever field appears last (and I believe it parses both fields) instead of preferring `startTimeStr`. Now since we are likely going to opt for a standardized format for the schedules, this may not matter, but worth keeping in mind this slight loss of functionality.

bryantgeorge (approved):

client-course-schedulizer/src/utilities/helpers/readCSV.ts
Jdellis818: No. It will need to be added back once we can parse duration from `math-schedule.csv`. I was struggling with that so I think it should be in a future PR

Jdellis818: I guess I'll add it back so we don't forget later though

charkour (approved):
Yeah, the changes look good. I noticed a todo on parsing a couple things, but before we creating parsing function for those, we should sit down with the registrar to figure out the required schedule and data formats. 

Also, I'd recommend using Moment.js to handle anything that is date or time related. 

I say we merge this in and these future comments can be addressed in another PR.

PR #29, opened 2020-11-07
Add teaching loads table

### LINKED ISSUE #30: Add Teaching Loads Table

Jdellis818:
This adds a table to the Teaching Loads tab. It's in a similar format to Prof. Vander Linden's spreadsheet. `Other Duties`, `Other Hours`, and `Load Notes` still need to be added to the table, but I think they should be in a different PR since we need to think more about how those will be added and saved.

### DIFF
--- a/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
+++ b/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-key */
 import {
   Paper,
   Table,
@@ -5,14 +6,14 @@ import {
   TableCell,
   TableContainer,
   TableHead,
-  TableRow,
+  TableRow as FacultyRow,
 } from "@material-ui/core";
 import React, { useContext, useMemo } from "react";
 import { Column, useTable } from "react-table";
-import { Schedule, Term } from "../../../utilities/interfaces/dataInterfaces";
+import { Course, Schedule, Section, Term } from "../../../utilities/interfaces/dataInterfaces";
 import { ScheduleContext } from "../../../utilities/services/context";
 
-interface TableData {
+interface FacultyRow {
   faculty: string;
   fallCourseSections?: string;
   fallHours?: number;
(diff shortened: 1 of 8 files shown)


client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
```diff
@@ -0,0 +1,173 @@
...
+  otherDuties?: string;
+  otherHours?: number;
+  springCourseSections?: string;
+  springHours?: number;
+  summerCourseSections?: string;
+  summerHours?: number;
+  totalHours?: number;
+}
+
+const createTable = (schedule: Schedule): TableData[] => {
```
charkour: Perhaps make a better name for the interface since it is an array? I guess I'm just a little confused as to if it is an individual row of data or data for multiple tables.

client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const newTableData: TableData[] = [];
+  schedule.courses.forEach((course) => {
+    course.sections.forEach((section) => {
+      section.instructors.forEach((instructor) => {
+        const instructorFullName = `${instructor.firstName} ${instructor.lastName}`;
+        const sectionName = `${course.prefixes[0]}-${course.number}-${section.letter}`;
+        const newDataRow: TableData = {
+          faculty: instructorFullName,
+        };
+        const [oldDataRow] = newTableData.filter((data) => {
```
charkour: A little confused by this. Is this saying that the first element of newTableData.filter() is the oldDataRow as in the previous row of data? Am I understanding that incorrectly?

client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
```diff
@@ -0,0 +1,173 @@
...
+              prepareRow(row);
+              return (
+                // Apply the row props
+                <TableRow {...row.getRowProps()} key={row.id}>
+                  {
+                    // Loop over the rows cells
+                    row.cells.map((cell) => {
+                      // Apply the cell props
+                      return (
+                        <TableCell {...cell.getCellProps()} key={cell.value}>
```
charkour: Not sure which key is giving duplicates, but this one potentially will give duplicates. You'll have to find a unique value for the key, perhaps stringifying the JSON object like `JSON.stringify(cell)` would guarantee uniqueness. Unique keys are important for optimized updating to the DOM.

client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
```diff
@@ -0,0 +1,173 @@
...
+        } else {
+          newTableData.push(newDataRow);
+        }
+      });
+    });
+  });
+  return newTableData.map((row) => {
+    return {
+      ...row,
+      totalHours: (row.fallHours || 0) + (row.springHours || 0) + (row.summerHours || 0),
```
charkour: Nice way to check for 0s!

client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
```diff
@@ -0,0 +1,173 @@
...
+      section.instructors.forEach((instructor) => {
+        const instructorFullName = `${instructor.firstName} ${instructor.lastName}`;
+        const sectionName = `${course.prefixes[0]}-${course.number}-${section.letter}`;
+        const newDataRow: TableData = {
+          faculty: instructorFullName,
+        };
+        const [oldDataRow] = newTableData.filter((data) => {
+          return data.faculty === instructorFullName;
+        });
+        switch (section.term) {
```
charkour, line 85: Better way for this switch? Lots of similar repetitions but I'm not if it can be better. Maybe the cases should stay but turn this code below into a function. If you find a bug in one, then you would have to fix it in all. 
```
            newDataRow.fallCourseSections = oldDataRow?.fallCourseSections
              ? `${oldDataRow.fallCourseSections}, ${sectionName}`
              : sectionName;
            newDataRow.fallHours = oldDataRow?.fallHours
              ? oldDataRow.fallHours + (section.facultyHours || course.facultyHours)
              : section.facultyHours || course.facultyHours;
```

charkour (changes requested):
This looks really really good. I just have a couple of concerns about readability. Using react-table was a great call and I loved the high-level TS and React you were using: quite impressive. I left a couple comments you should resolve before approval; this is a great start. Please let me know if you have any questions of I you think my suggestions are wrong (which has been the case before). 

As for the updated type, I created union types because they all depend on each other. For example, all of the hours fields should be handled the same way, and in this case, they are optional numbers. But if we change how the hours are handled, then we just update them in one place. Lmk if using union types is dumb; I wanted to try it out. I learned it from [this video](https://www.youtube.com/watch?v=RjQpep8fBdo&list=LL&index=4). Thanks.

client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
Jdellis818: Yeah I thought about doing that but decided that it may be confusing with the number of parameters that need to be passed, and assignment in a different function is kind of weird, but you're right that it's probably better since it'd prevent the duplicated code so I'll do my best to fix that

client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
Jdellis818: No it's a data row that's been updated previously because a section has already been added for that professor. I'll try to think of a better name.

client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
Jdellis818: It's an individual row so I'll change the name to TableRow

client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
Jdellis818: I'm realizing that this issue is just eslint being dumb. `getCellProps()` and the other getter methods to get the props returns a unique key prop, so it shouldn't be necessary to specify a key. Is it alright to just disable that eslint error for this file?

charkour: @Jonri2, yup, it's fine to disable that key rule for the file. I still don't know what's best for ESLint. Also, you could call `const cellProps = cell.getCellProps()` and then do `<TabelCell key={cellProps.key} {...cellProps}>` if you prefer this way.

charkour (approved):
Thanks for making these changes! It's easier to read and maintain. LGTM and you can merge it when you're ready.

charkour (approved):
Good catch on that bug. I missed that. I guess that's a good example of how important readability is. 

Anyways, this looks great. I'm impressed this works with the entire Course Section Encrollment.csv . Since we have the official document format, I think the next step is to make unit tests for this.

PR #32, opened 2020-11-08
refactor schedule

### LINKED ISSUE #31: Refactor Schedule Component

### LINKED ISSUE #35: Calling setState in a useEffect
image

This is reproduced by quickly uploading multiple different spreadsheets.

charkour:
I started to work on getting CSV content to show on the Faculty Schedule and realized I first wanted to refactor the Schedule component to make it reusable (for the Room Schedule) also. This PR adds some small styling fixes to the Schedule header and left-axis. It also includes a fakeSchedule object to use when developing so you don't need to constantly import the CSV. Schedule headers now are able to take data and FacultySchedule uses context to do so. There is also conditional rendering on the app to say when there is no schedule. 

I added some TODOs to address in the future. 

To test, `npm i` and then `npm start` and upload a CSV to see the updates to the FacultySchedule. Please let me know your thoughts! This is my best for now but I know I overlooked things.

closes #31 

**UPDATE** see below


bryantgeorge (commented):
Overall, looks good, but with the "No schedule selected...", I wonder if this could be annoying to a user that wants to build a schedule from scratch on our app without a base CSV (maybe an uncommon scenario?). In any case, it might be nice to have the add a section button still visible even with no schedule CSV loaded in?

charkour:
That is a great point. I'll make that update. Thanks!

client-course-schedulizer/src/utilities/services/facultySchedule.ts
```diff
@@ -0,0 +1,16 @@
+import { Schedule } from "../interfaces/dataInterfaces";
+
+// Get list of unique professors.
+export const getProfs = (schedule: Schedule): string[] => {
+  const professors: string[] = [];
+  schedule.courses.map((course) => {
+    return course.sections.map((section) => {
```
Jdellis818: Change these to `forEach` since you're not doing anything with the returned array from `map`

client-course-schedulizer/src/utilities/services/schedule.ts
```diff
@@ -0,0 +1,11 @@
+// Returns a list of hours to display on the Schedule
+// TODO: add better types for timing, maybe: https://stackoverflow.com/questions/51445767/how-to-define-a-regex-matched-string-type-in-typescript
+export const getHoursArr = (min: string, max: string): number[] => {
+  const minHour = parseInt(min.split(":")[0]);
+  const maxHour = parseInt(max.split(":")[0]);
+  const hours: number[] = [];
+  for (let i = minHour; i < maxHour; i += 1) {
+    hours.push(i);
+  }
```
Jdellis818: I'm not sure if your way is better, but I think this would be the functional way to get all integers in a range. It's pretty unreadable though, so you can decide which way you prefer.
https://stackoverflow.com/questions/8069315/create-array-of-all-integers-between-two-numbers-inclusive-in-javascript-jquer/8069367
```
const hours = Array(maxHour - minHour + 1).fill().map((_, idx) => minHour + idx)
```

Jdellis818 (changes requested):
As a whole, I like your updates. Being able to dynamically set the professors and hours is really good. The only issue is that the app crashes if you try to import a CSV multiple times saying `Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`. I worked on trying to fix the bug, but couldn't find the source. It says it's coming from `Stick` but that could just be because of how we're updating `professors` and `schedule`.

client-course-schedulizer/src/utilities/services/schedule.ts
charkour: I like the functional way. Thanks!

client-course-schedulizer/src/utilities/services/facultySchedule.ts
charkour: Good call. Thanks

charkour:
Okay, there are a lot of changes going on here now. I'll try to explain what's going on so you can accurately review it. 

1. I added lodash to have better and more performant methods to manipulating arrays and objects.
1. Fixed the bug that was calling a setState inside a useEffect. More on that later. (#35) 
1. Created a reducer to manage dependant global states
1. Changed how the ImportButton responds to importing.
1. Added an async component that responds to a loading state
1. Added more functionality to the AddSectionButton
1. Added the flag `downlevelIteration` to true while compiling. See [this SO post](https://stackoverflow.com/questions/49218765/typescript-and-iterator-type-iterableiteratort-is-not-an-array-type)

2) So there was a bug with the Stick and StickyNode components that related to how we were setting the schedule once it was uploaded. I couldn't figure out but there was some sort of circular re-rendering, so I decided to remove the sticky components (all on the schedule) while the CSV was parsing. I didn't realize the reading the file in is an async function, but setting a loading state based on loading the file seemed to fix this. Please try to break it! Right now, `isLoading` is in the global context, maybe it should be `isCSVLoading` and we should keep it in the global context since the whole app depends on it. 

3) When we update the schedule in context, we also want it to update the professors (and other things in the future, like rooms), so reducer is a way to update multiple states at once that depend on each other. I placed the reducer in the global context so it can be accessed anywhere in the DOM. This took me awhile to understand so I would be happy to sit down with you to talk through how it works! Just let me know.

Let me know what you think and if there are any changes! I think fixing the bug makes the schedule headers less laggy.

client-course-schedulizer/src/components/Header/ImportButton/ImportButton.tsx
```diff
@@ -1,28 +1,35 @@
...
 
     read.onloadend = () => {
       const scheduleJSON = readCSV.csvStringToSchedule(String(read.result));
-      setSchedule(scheduleJSON);
-      // eslint-disable-next-line no-console
-      console.log(scheduleJSON);
+      // TODO: store in local storage incase prof navigates away while editing.
+      // currently a redundant check
+      if (!isEqual(schedule, scheduleJSON)) {
+        appDispatch({ payload: { schedule: scheduleJSON }, type: "setScheduleData" });
```
Jdellis818, line 29: Woah this is kinda intense. Super cool!

client-course-schedulizer/src/components/reuseables/AsyncComponent/AsyncComponent.tsx
```diff
@@ -0,0 +1,12 @@
...
+interface AsyncComponent {
+  isLoading: boolean;
+}
+
+/*
+  Show a loading state while waiting on a async function to finish.
+*/
+export const AsyncComponent = ({ isLoading, children }: PropsWithChildren<AsyncComponent>) => {
+  return <>{isLoading ? <div>parsing csv...</div> : <>{children}</>}</>;
+};
```
Jdellis818, line 12: This is really cool. Well done!

client-course-schedulizer/src/utilities/services/facultySchedule.ts
```diff
@@ -0,0 +1,15 @@
+import forEach from "lodash/forEach";
+import { Schedule } from "../interfaces/dataInterfaces";
+
+// Get list of unique professors.
+export const getProfs = (schedule: Schedule): string[] => {
+  const professorsSet = new Set<string>();
```
Jdellis818, line 6: Good idea to make this a Set.

client-course-schedulizer/src/utilities/services/schedule.ts
```diff
@@ -0,0 +1,9 @@
+import range from "lodash/range";
+
+// Returns a list of hours to display on the Schedule
+// TODO: add better types for timing, maybe: https://stackoverflow.com/questions/51445767/how-to-define-a-regex-matched-string-type-in-typescript
+export const getHoursArr = (min: string, max: string): number[] => {
+  const minHour = parseInt(min.split(":")[0]);
+  const maxHour = parseInt(max.split(":")[0]);
+  return range(minHour, maxHour);
+};
```
Jdellis818, line 9: Nice! This is way better than before

Jdellis818 (approved):
This LGTM. The reducer stuff definitely went over my head, but it seems to work and it looks right. Thanks for working so hard to make the app work well!

client-course-schedulizer/src/components/Header/ImportButton/ImportButton.tsx
charkour: Haha thanks! I think you'll really like this once you get the hang of it.

PR #34, opened 2020-11-08
Export converted csv

### LINKED ISSUE #20: Export as CSV

bryantgeorge:
This branch:
1. Makes the import button only parse the CSV into a `scheduleJSON`, save the `scheduleJSON` in the context, and log it to the console (ie. it no longer logs the JSON converted back to CSV).
2. Makes the export button work (ie. it will save a file called `schedule.csv` to your downloads with the context's `scheduleJSON` converted back into a CSV).
3. Makes the hamburger menu open to the right of the hamburger button (such that you can close the menu by clicking the hamburger button again), makes the import and export menu item buttons clickable for the entire width and height of the menu item (previously only the text portion was clickable, so there was a boarder that looked clickable but wouldn't trigger anything), and finally makes the buttons have the click icon when hovered (previously the label would only give the default mouse icon when hovered).

To Test:
1. `npm install` and `npm start` like usual.
2. Try out importing a CSV, and then exporting it.
3. Make sure that the changes to the import/export button styling make sense/look nice.

Closes #20


client-course-schedulizer/src/components/Header/ExportButton/ExportButton.scss
```diff
@@ -0,0 +1,15 @@
...
+.hidden {
+  display: none !important;
+}
+
+.export-label {
+  padding-top: 6px !important;
+  padding-bottom: 6px !important;
+  padding-left: 16px !important;
+  padding-right: 16px !important;
+  cursor: pointer !important;
```
Jdellis818: Most of these styles need `!important` to override the MUI default styles, but I think you can remove it for the cursor

client-course-schedulizer/src/components/Header/ExportButton/ExportButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+  element.setAttribute("href", `data:text/plain;charset=utf-8,${encodeURIComponent(text)}`);
+  element.setAttribute("download", filename);
+
+  element.style.display = "none";
+  document.body.appendChild(element);
+
+  element.click();
+
+  document.body.removeChild(element);
+};
```
Jdellis818: Maybe use [this package](https://www.npmjs.com/package/js-file-download) to implement this functionality? I think it does exactly the same thing but provides cross-browser support.

Jdellis818 (changes requested):
This looks really good to me and it works really well! There are just a couple small changes to address. Most importantly, I think it'd be good to use a package for the download function so that it works for all browsers and decreases the amount of work for us. Great work though! Your code looks really good

client-course-schedulizer/src/components/Header/ExportButton/ExportButton.tsx
charkour: I agree with Jon. I did some digging and [found the package does something similar](https://github.com/kennethjiang/js-file-download/blob/master/file-download.js) with a couple of extra features to support Safari and IE which is cool!

client-course-schedulizer/src/components/Header/ExportButton/ExportButton.scss
charkour: Additionally, I suggest combining both of the `.export-label { ... }` rule sets. The styles look good.

client-course-schedulizer/src/components/Header/ImportButton/ImportButton.scss
```diff
@@ -5,3 +5,11 @@
...
   display: none !important;
 }
+
+.import-label {
+  padding-top: 6px !important;
+  padding-bottom: 6px !important;
+  padding-left: 16px !important;
+  padding-right: 16px !important;
+  cursor: pointer !important;
+}
```
charkour: Since these are the same styles as the export button, I would combine both of the styles and put the styles in an `.scss` file for the parent component (I think header.scss or something similar). Then you could have only one .hidden class and then `.import-label, .export-label { ... }`.

charkour (changes requested):
Looks good! Just a couple of things to address to reduce repeated code. Nice work.

client-course-schedulizer/src/components/Header/ImportButton/ImportButton.scss
bryantgeorge: 7a9bda1

client-course-schedulizer/src/components/Header/ExportButton/ExportButton.tsx
bryantgeorge: Thanks, this package makes things simpler. I tried IE, but the site didn't seem to load at all (but I guess IE isn't that important to support). With Safari, I don't have a great way to test Safari, so if either of you do that would be greatly appreciated.

7a9bda1

client-course-schedulizer/src/components/Header/ExportButton/ExportButton.scss
bryantgeorge: 7a9bda1

Jdellis818 (approved):
LGTM once the conflicts are resolved

charkour (approved):
LGTM! Thanks for the awesome work.

bryantgeorge:
Conflicts resolved, merging now

PR #36, opened 2020-11-11
Add Unit Tests

### LINKED ISSUE #33: Add Unit Testing to Import Function

Jdellis818:
This PR adds unit tests for the CSV to Schedule conversion based on `math-schedule.csv`. It also adds parsing for `semesterLength` and `duration` for spreadsheets in the registrar's format. To run the tests simply use the command `npm run test`. I also removed the Create React App icons from the manifest so we'd stop getting that console error every time our app loads.


bryantgeorge (approved):
LGTM, tests pass and seem reasonable. Good to see those icons removed to reduce the console errors. Nice use of the moment package, that should make our date parsing code much more readable.

I am curious though given this new schedule format, we will have to change the export function to match the registrar format, and I don't know if certain fields will be reconstruct-able (things like instruction method, term start/end, etc.)

charkour (approved):
LGTM! I agree with @Syobnaf about how we may need to update the export function to match the correct format. But that can wait for another day.

Jdellis818:
> I am curious though given this new schedule format, we will have to change the export function to match the registrar format, and I don't know if certain fields will be reconstruct-able (things like instruction method, term start/end, etc.)

True. We might just need to store those values in the object as well

PR #40, opened 2020-11-13
Add events to calendar

### LINKED ISSUE #37: Add Events to Calendar

Jdellis818:
This PR adds events to the calendars and allows you to scroll through semesters. It does so by creating a `GroupedEvents` objects which groups events by the faculty member and then loads the events on their calendar.

### DIFF
--- a/client-course-schedulizer/src/components/Toolbar/SemesterSelector/SemesterSelector.tsx
+++ b/client-course-schedulizer/src/components/Toolbar/SemesterSelector/SemesterSelector.tsx
@@ -1,16 +1,43 @@
 import { IconButton, Typography } from "@material-ui/core";
 import { ChevronLeft, ChevronRight } from "@material-ui/icons";
-import React from "react";
+import React, { useContext } from "react";
+import { Term } from "../../../utilities/interfaces/dataInterfaces";
+import { AppContext } from "../../../utilities/services/appContext";
 import "./SemesterSelector.scss";
 
 export const SemesterSelector = () => {
+  const terms: Term[] = Object.keys(Term).map((term) => {
+    return Term[term as keyof typeof Term];
+  });
+  const {
+    appState: { selectedTerm },
+    appDispatch,
+  } = useContext(AppContext);
   return (
     <div className="semester-selector">
-      <IconButton>
+      <IconButton
+        disabled={selectedTerm === terms[0]}
+        onClick={() => {
+          const newTerm = terms[terms.indexOf(selectedTerm) - 1];
+          appDispatch({
+            payload: { term: newTerm },
+            type: "setSelectedTerm",
+          });
+        }}
+      >
         <ChevronLeft />
       </IconButton>
-      <Typography variant="h6">Fall 2021</Typography>
-      <IconButton>
+      <Typography variant="h6">{selectedTerm}</Typography>
+      <IconButton
+        disabled={selectedTerm === terms[terms.length - 1]}
+        onClick={() => {
+          const newTerm = terms[terms.indexOf(selectedTerm) + 1];
+          appDispatch({
+            payload: { term: newTerm },
+            type: "setSelectedTerm",
+          });
+        }}
+      >
         <ChevronRight />
       </IconButton>
     </div>
--- a/client-course-schedulizer/src/components/Toolbar/SemesterSelector/SemesterSelector.tsx
+++ b/client-course-schedulizer/src/components/Toolbar/SemesterSelector/SemesterSelector.tsx
@@ -1,42 +1,40 @@
 import { IconButton, Typography } from "@material-ui/core";
 import { ChevronLeft, ChevronRight } from "@material-ui/icons";
 import React, { useContext } from "react";
+import { enumArray } from "../../../utilities/helpers/utils";
 import { Term } from "../../../utilities/interfaces/dataInterfaces";
 import { AppContext } from "../../../utilities/services/appContext";
 import "./SemesterSelector.scss";
 
 export const SemesterSelector = () => {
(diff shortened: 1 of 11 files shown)


client-course-schedulizer/src/components/Toolbar/SemesterSelector/SemesterSelector.tsx
```diff
@@ -1,16 +1,43 @@
...
     <div className="semester-selector">
-      <IconButton>
+      <IconButton
+        disabled={selectedTerm === terms[0]}
+        onClick={() => {
+          const newTerm = terms[terms.indexOf(selectedTerm) - 1];
+          appDispatch({
+            payload: { term: newTerm },
+            type: "setSelectedTerm",
+          });
```
charkour: This onClick handler I would make into a function take takes an index parameter since it is used here and below. If we update the logic in one, it'll need to be updated in both.

client-course-schedulizer/src/components/Toolbar/SemesterSelector/SemesterSelector.tsx
```diff
@@ -1,16 +1,43 @@
...
 import { ChevronLeft, ChevronRight } from "@material-ui/icons";
-import React from "react";
+import React, { useContext } from "react";
+import { Term } from "../../../utilities/interfaces/dataInterfaces";
+import { AppContext } from "../../../utilities/services/appContext";
 import "./SemesterSelector.scss";
 
 export const SemesterSelector = () => {
+  const terms: Term[] = Object.keys(Term).map((term) => {
+    return Term[term as keyof typeof Term];
```
charkour: Wow, I didn't know this was possible. Iterating over enumerations in typescript is kinda weird. haha nice job

client-course-schedulizer/src/utilities/services/schedule.ts
```diff
@@ -7,3 +12,67 @@ export const getHoursArr = (min: string, max: string): number[] => {
...
+export interface GroupedEvents {
+  [key: string]: EventInput[];
+}
+
+// TODO: Add events with no meeting times as all day
+export const getEvents = (schedule: Schedule, groups: "faculty" | "room"): GroupedEvents => {
+  const events: GroupedEvents = {};
+  const days: Day[] = Object.keys(Day).map((day) => {
+    return Day[day as keyof typeof Day];
+  });
```
charkour: I've seen this a couple places. Would it be better to make a function to do this? 

```
const enumArray = <T>(e: T) => {
  return Object.keys(e).map((val) => {
    return e[val as keyof typeof e];
  });
};
```

charkour: use: `const days: Day[] = enumArray(Day);`

client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx
```diff
@@ -1,37 +1,67 @@
...
+  } = useContext(AppContext);
+
   const times = {
-    slotMaxTime: calendarOptions.slotMaxTime as string,
-    slotMinTime: calendarOptions.slotMinTime as string,
+    slotMaxTime,
+    slotMinTime,
+  };
+
+  calendarOptions = {
```
charkour, line 37: I would just say `// add context times to the existing calendarOptions`

client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx
```diff
@@ -1,37 +1,67 @@
...
+    groupedEvents[key] = filter(groupedEvents[key], (e) => {
+      return e.extendedProps?.section.term === selectedTerm;
+    });
+  });
+
+  // Filter out headers with no events
+  const calendarHeadersNoEmptyInTerm = filter(calendarHeaders, (header) => {
+    const groupEvents = groupedEvents[header];
+    return groupEvents?.length > 0;
+  });
```
charkour: It might be worth doing research to see if it is important that we move logic out of the function component. I think these get redefined on every render, but maybe we put these in a useCallback()? Not sure. would be worth researching.

charkour (changes requested):
This is some great work! I like how seemlessly it works together and seeing the calendars with the events is awesome. I like to use of moment and also how you wrote code that is extensible and flexible. I think it would be worth to add some unit tests in the future to test the sorting and filtering logic. 

A couple small things to address before merging.

client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx
Jdellis818: I changed this to `useMemo`, but I'm unsure whether the dependencies do anything since they include objects and I'm worried that it does a reference comparison rather than a value comparison. If it uses reference comparison then I think the code would re-run on any re-render since an object would have a new location in memory even if it had the same values.

Jdellis818: I did some research though and it seems like it may not be worth worrying about anyway since the operation is not computationally expensive. React is fast enough that a re-render isn't a big deal

charkour (approved):
I did some preliminary research on `useMemo()` in our app and counted the number of re-renderes. When `useMemo()` is not used, `forOwn()` is called 15 times and `calendarHeadersNoEmptyInTerm()` is called 15 times as well; however, with the memoization, `filteredEvents` and `calendarHeadersNoEmptyInTerm` are only redefined twice each. I think getting these called twice is fine! In my experience, it is really hard to get things called twice by hooks to be called only once since there are many complexities that I do not understand. 

I liked the rest of the updates. You can merge this when ready!

PR #46, opened 2020-11-17
Feature/start room schedule

### LINKED ISSUE #44: Add Room Schedule

### LINKED ISSUE #45: calling setState in a useEffect when switching tabs on the Room Schedule
image

charkour:
*UPDATES*
------
- Fixed the wording to make more sense when there is no courses
- Fixed and over-engineered a system of creating subcomponents. My solution I don't think is super performant so maybe we will realize that in the future. 
- Awaited all appDispatch so see what happens. 

Also, I'm sorry that this PR is bigger than it should usually be. 

Original
------

- This PR adds the Room Schedule view. 
- It also fixes a bug that causes the app to crash on excessive `setState` calls inside a `useEffect` by awaiting the `appDispatch` action called by `<SemesterSelector>`. 
- Also, it adds functions to the buttons when there is no CSV uploaded. 
- It adds context to each Schedule.
- It adds subcomponents to `<AsyncComponent>` 

To test, run the app, and upload a CSV. 

Please let me know what you think! Your input is great. 

closes #44 
fixes #45


client-course-schedulizer/src/components/Header/ImportButton/ImportButton.tsx
```diff
@@ -1,48 +1,32 @@
...
-      // TODO: store in local storage incase prof navigates away while editing.
-      // currently a redundant check
-      if (!isEqual(schedule, scheduleJSON)) {
-        appDispatch({ payload: { schedule: scheduleJSON }, type: "setScheduleData" });
-      }
-      setIsLoading(false);
-    };
-  };
+export const ImportButton = ({ className }: ImportButton) => {
+  const onInputChange = useImportFile();
```
Jdellis818, line 10: I like the use of a custom hook here!

client-course-schedulizer/src/components/Tabs/Tabs.tsx
```diff
@@ -37,52 +38,53 @@ export const Tabs = () => {
...
-          )}
-        </>
-      </AsyncComponent>
-    </Paper>
+    <AsyncComponent isLoading={isCSVLoading}>
+      <AsyncComponent.Loading>Parsing CSV...</AsyncComponent.Loading>
+      <AsyncComponent.Loaded>
+        {schedule.courses.length === 0 ? (
+          <>
+            <h2>No schedule selected. Please import a CSV or add a section to start Editing.</h2>
```
Jdellis818: I think it may be clearer to say "No schedule selected. Please import a CSV of an existing schedule or start building your schedule from scratch by adding a section."

Jdellis818: This wording is still a little unclear so feel free to edit it, but I think I would be confused what exactly each of the buttons do when I reach the landing page.

client-course-schedulizer/src/components/Toolbar/SemesterSelector/SemesterSelector.tsx
```diff
@@ -12,14 +13,19 @@ export const SemesterSelector = () => {
...
 
   const handleOnClick = (index: number) => {
-    return () => {
+    return async () => {
+      setIsScheduleLoading(true);
       const newTerm = terms[index];
-      appDispatch({
+      // This action takes so long it affectively makes this
+      //  synchronous function asynchronous.
+      await appDispatch({
```
Jdellis818, line 24: Sorry for missing that we needed this! Thanks for fixing it

client-course-schedulizer/src/components/reuseables/AsyncComponent/AsyncComponent.tsx
```diff
@@ -1,12 +1,71 @@
...
+}: PropsWithChildren<AsyncComponent>) => {
+  // TODO: better way to get all of the children? Loop?
+  const loadedNode = getSubComponent(children, "Loaded");
+  const loadingNode = getSubComponent(children, "Loading");
+
+  return <>{isLoading ? <>{loadingNode}</> : <>{loadedNode}</>}</>;
+};
+
+const Loading: FC<{}> = ({ children }: PropsWithChildren<{}>): JSX.Element => {
+  return <>{children}</>;
```
Jdellis818: Woah...

client-course-schedulizer/src/utilities/hooks/useImportFile.ts
```diff
@@ -0,0 +1,35 @@
...
+    const file: Blob | null = e.target.files && e.target.files[0];
+    const read = new FileReader();
+    file && read.readAsBinaryString(file);
+
+    read.onloadend = () => {
+      const scheduleJSON = readCSV.csvStringToSchedule(String(read.result));
+      // TODO: store in local storage incase prof navigates away while editing.
+      // currently a redundant check
+      if (!isEqual(schedule, scheduleJSON)) {
+        appDispatch({ payload: { schedule: scheduleJSON }, type: "setScheduleData" });
```
Jdellis818: Is there a reason we don't have to await appDispatch here? I guess I'm still a little confused about what the issue was before

Jdellis818 (approved):
LGTM. It seems to work well. The only thing I would change is the text on the landing screen to be a little more clear about what the options are to start using the app. The Async Component stuff went way over my head haha, but it seems really cool.

Jdellis818:
I just noticed that if there is a section with more than 1 prof, it's added as 2 sections at the same time in the same room on the room schedule. This is not a bug I think we have to worry about on this PR, but it's something to consider in the future. (This is an issue with my code, not yours)

client-course-schedulizer/src/utilities/hooks/useImportFile.ts
charkour: Yeah, I'm not sure. Maybe we could just await every `appDispatch` call just to be safe? I didn't think it would be an issue but it has something to do with how massive the schedule is. 

I don't understand the behavior. But I can add the await's everywhere and see what happens.

client-course-schedulizer/src/components/reuseables/AsyncComponent/AsyncComponent.tsx
charkour: Haha, I'll take this as a note to add better documentation for this. The typings make it seem extra complicated, but it's really not.

client-course-schedulizer/src/components/Tabs/Tabs.tsx
charkour: Good call. Thanks

Jdellis818 (approved):
LGTM

Issue #43, opened 2020-11-16
Allow for Multiple Meetings


bryantgeorge:
Let newlines in meeting fields be converted into multiple meetings

Issue #39, opened 2020-11-13
Improve Export Function


bryantgeorge:
Involves:
1. Making the exported CSV conform to the registrar's format (may need to add some fields to our interfaces).
2. Looking into bugs like partial rows and commas within fields.

PR #47, opened 2020-11-18
Add sections to context

bryantgeorge:
This branch will add courses/sections to the schedule when using the add a course form.
- `npm install` and `npm start`.
- Try uploading a CSV, then use the form to add a new course.
- Export to see that it appears (and is not duplicated since this was something I ran into, and I think I fixed it, but check to be sure).
- Also try adding a new section to an existing course (may need to console log the schedule before and after to see that it reuses the existing course).
- I disabled the ability to save a course/section using the enter key because I found it annoying when I was using the keyboard to navigate the form, and I would click enter instinctively which would then save an incomplete course/section.
- Branch also allows for multiple instructors, prefixes, etc. to be specified by utilizing the case functions that the parser uses (so separate instructors by `;` or `,`). This is why the case functions have been separated into callbacks and case functions (such that they are more general and usable elsewhere).

I tried for a long time to make a cool auto-fill drop-down to work for things like instructors (see https://codesandbox.io/s/autocomplete-freesolo-with-auto-value-on-blur-e2smn?file=/demo.js), but for the life of me I could not figure it out because there was a strange bug where clicking an option would result in a 0 being returned instead of the instructor's name :(

Note that some stuff in the export may look strange still, but hopefully a lot of this will be improved in #39 and #43 

Closes #47


client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
```diff
@@ -24,175 +24,227 @@ const thursReg = RegExp("[Tt][Hh]|[Rr]");
...
-  } else if (sectionLength > 35 && sectionLength <= 80) {
-    section.semesterLength = firstStartMonths.includes(startMonth)
-      ? di.SemesterLength.HalfFirst
-      : di.SemesterLength.HalfSecond;
-  } else {
-    // TODO: Figure out if intensive is A, B, C, or D
-    section.semesterLength = di.SemesterLength.IntensiveA;
+    return SemesterLength.Full;
   }
+  if (sectionLength > 35 && sectionLength <= 80) {
```
charkour: It could be good to make these numbers be constants. Without the context, it is hard to derive the significance of 35 and 80. I know I've forgotten to do that too so you can remind me to do so when I forget. 

something like `const MIN_HALF_LENGTH = 35;`

charkour (changes requested):
This looks great! I love how it works and shows up in context. Also, way to rework the case functions to reuse the logic. 

Only one comment to address.

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -104,53 +121,60 @@ export const AddSectionPopover = () => {
...
+      sections: [],
       studentHours: Number(data.studentHours),
     };
-    // eslint-disable-next-line no-console
-    return console.log(newCourse);
+
+    // Insert the Section to the Schedule, either as a new Course or to an existing Course
+    insertSectionCourse(schedule, newSection, newCourse);
+
+    appDispatch({ payload: { schedule }, type: "setScheduleData" });
```
Jdellis818, line 149: Maybe await this dispatch? I know we had an issue with not awaiting on a different PR, but I'm still confused as to why

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
Jdellis818: My bad... I believe I made these magic numbers in a previous PR haha

Jdellis818 (changes requested):
LGTM

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
charkour: Haha oops, I missed them too.

Jdellis818 (approved):

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
bryantgeorge: When I tried this (like `await appDispatch({ payload: { schedule }, type: "setScheduleData" });`), I got a warning saying: `'await' has no effect on the type of this expression. ts(80007)`, so I guess this isn't necessary? Feel free to correct me if I'm misunderstanding.

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
bryantgeorge: f14e906

Jdellis818 (approved):
LGTM

charkour (approved):
LGTM. You'll just need to change it from a draft.

bryantgeorge:
> LGTM. You'll just need to change it from a draft.

Oh thanks I forgot that here and on my other PR, oops

PR #50, opened 2020-11-20
Allow multiple meetings

### LINKED ISSUE #43: Allow for Multiple Meetings

bryantgeorge:
(Probably should wait to review this until #47 is merged to `develop` because otherwise the changes from that branch will appear in this branch's `Files changed` tab)

To Test:
- `npm install` and `npm start`.
- First try importing and exporting the `math-schedule.csv`, and notice that the exported meetings for classes with multiple meetings are on multiple lines as expected (includes days, startTimes, locations, roomCapacities, durations).
- Also note that the bug with commas in the name of a course is fixed (should only appear in one field), but this could still be a problem with other fields...
- Now try using the add a course/section popup to add a new meeting on an existing section, so for example, try Math 100 A with new days, duration, location, roomCapacity, startTime, and notice that it is added to the same section when exporting.
- Try to break it with edge cases?

Might be useful to log the schedule after parsing/adding courses/sections/meetings for testing purposes.

### DIFF
--- a/client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
+++ b/client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
@@ -27,10 +27,9 @@ const thursReg = RegExp("[Tt][Hh]|[Rr]");
 const friReg = RegExp("[Ff]");
 const satReg = RegExp("[Ss](?![Uu])");
 
-export const startTimeCallback = (value: string, { meetings }: CaseCallbackParams) => {
-  const startTimes = value.split("\n");
-  let i = 0;
-  startTimes.forEach((startTime) => {
+const createMeetings = (value: string, { meetings }: CaseCallbackParams): string[] => {
+  const valueParts = value.split("\n");
+  for (let i = 0; i < valueParts.length; i += 1) {
     if (meetings.length <= i) {
       // If there aren't enough meetings, create a new one
       meetings.push({
@@ -40,26 +39,23 @@ export const startTimeCallback = (value: string, { meetings }: CaseCallbackParam
         startTime: "",
       });
     }
+  }
+  return valueParts;
+};
+
+export const startTimeCallback = (value: string, params: CaseCallbackParams) => {
+  const startTimes = createMeetings(value, params);
+  const { meetings } = params;
+  startTimes.forEach((startTime, i) => {
     meetings[i].startTime = startTimeCase(startTime);
-    i += 1;
   });
 };
 
-export const locationCallback = (value: string, { meetings }: CaseCallbackParams) => {
-  const locations = value.split("\n");
-  let i = 0;
-  locations.forEach((location) => {
-    if (meetings.length <= i) {
-      // If there aren't enough meetings, create a new one
-      meetings.push({
-        days: [],
-        duration: 0,
-        location: { building: "", roomCapacity: 0, roomNumber: "" },
-        startTime: "",
-      });
-    }
+export const locationCallback = (value: string, params: CaseCallbackParams) => {
+  const locations = createMeetings(value, params);
+  const { meetings } = params;
+  locations.forEach((location, i) => {
     [meetings[i].location.building, meetings[i].location.roomNumber] = locationCase(location);
-    i += 1;
   });
 };
 
@@ -71,21 +67,11 @@ export const semesterLengthCallback = (value: string, { section }: CaseCallbackP
   section.semesterLength = semesterLengthCase(value);
 };
 
-export const daysCallback = (value: string, { meetings }: CaseCallbackParams) => {
-  const daysStrings = value.split("\n");
-  let i = 0;
-  daysStrings.forEach((days) => {
-    if (meetings.length <= i) {
-      // If there aren't enough meetings, create a new one
-      meetings.push({
-        days: [],
-        duration: 0,
-        location: { building: "", roomCapacity: 0, roomNumber: "" },
-        startTime: "",
-      });
-    }
+export const daysCallback = (value: string, params: CaseCallbackParams) => {
+  const daysStrings = createMeetings(value, params);
+  const { meetings } = params;
+  daysStrings.forEach((days, i) => {
     meetings[i].days = daysCase(days);
-    i += 1;
   });
 };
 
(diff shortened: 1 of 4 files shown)


charkour:
I suggest you resolve the conflicts by merging in develop branch to this branch so it's easier to view the changes. Thanks! @Syobnaf

bryantgeorge:
> I suggest you resolve the conflicts by merging in develop branch to this branch so it's easier to view the changes. Thanks! @Syobnaf

Thanks for the heads up, should be good now.

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
```diff
@@ -27,12 +27,40 @@ const thursReg = RegExp("[Tt][Hh]|[Rr]");
 const friReg = RegExp("[Ff]");
 const satReg = RegExp("[Ss](?![Uu])");
 
-export const startTimeCallback = (value: string, { firstMeeting }: CaseCallbackParams) => {
-  firstMeeting.startTime = startTimeCase(value);
+export const startTimeCallback = (value: string, { meetings }: CaseCallbackParams) => {
+  const startTimes = value.split("\n");
+  let i = 0;
+  startTimes.forEach((startTime) => {
+    if (meetings.length <= i) {
```
Jdellis818: The `forEach` function has a second parameter for the index, so there's no need to keep track of it manually.

`startTimes.forEach((startTime, i) => { ... })`

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
```diff
@@ -27,12 +27,40 @@ const thursReg = RegExp("[Tt][Hh]|[Rr]");
...
+  locations.forEach((location) => {
+    if (meetings.length <= i) {
+      // If there aren't enough meetings, create a new one
+      meetings.push({
+        days: [],
+        duration: 0,
+        location: { building: "", roomCapacity: 0, roomNumber: "" },
+        startTime: "",
+      });
+    }
```
Jdellis818: This is repeated code, so it's probably best to make a function for it

Jdellis818: Also, replace the repeated code in the other callbacks with a call to the new function.

Jdellis818: You should put the `split()` line in the function too.

client-course-schedulizer/src/utilities/helpers/readCSV.ts
```diff
@@ -132,6 +126,21 @@ export const csvStringToSchedule = (csvString: string): Schedule => {
 };
 
 export const insertSectionCourse = (schedule: Schedule, section: Section, course: Course) => {
+  const { meetings } = section;
+
+  // Check if any meetings are empty, and should be removed
+  let i = 0;
+  let numRemoved = 0;
+  const meetingsCopy = meetings;
+  meetings.forEach((meeting) => {
```
Jdellis818: You can use the `forEach()` index parameter here too

client-course-schedulizer/src/utilities/helpers/readCSV.ts
```diff
@@ -132,6 +126,21 @@ export const csvStringToSchedule = (csvString: string): Schedule => {
...
+    if (meeting.days === [] || meeting.duration === 0) {
+      meetingsCopy.splice(i - numRemoved, 1);
```
Jdellis818: I think it would be more clear to do something like
```
const meetingIndex = meetingsCopy.indexOf(meeting);
meetingsCopy.splice(meetingsCopy.splice(meetingIndex, 1);
```
rather than keeping track of numRemoved, but I think that's just my personal preference and I'd be fine keeping it as is if you disagree.

Jdellis818: Also, I think this change may make it so you don't need to make a copy, but I'm not too sure.

Jdellis818 (changes requested):
I really like the fixed bugs and added functionality with allowing for multiple meetings! This kind of flexibility is the sort of thing that will help our users out immensely. I just had a few small changes to request to simplify some code. Also, the app seems to crash when I try to import the whole Course Section Enrollment CSV.

client-course-schedulizer/src/utilities/helpers/readCSV.ts
bryantgeorge: 325d1a9 and 172cce3

client-course-schedulizer/src/utilities/helpers/readCSV.ts
bryantgeorge: 325d1a9 and 172cce3 should make the logic here a lot simpler (don't know why I didn't think to use `filter()` before).

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
bryantgeorge: This is really good to know, thanks! cac67a9

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
bryantgeorge: I tried this in cac67a9, but I think in certain ways it makes things more confusing, so let me know if you want me to revert the change or have suggestions on how it could be done better.

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
```diff
@@ -27,12 +27,36 @@ const thursReg = RegExp("[Tt][Hh]|[Rr]");
...
+  }
```
charkour: I'd suggest making this for loop a forEach since it is similar to what we do elsewhere in the app. If you disagree, that's fine too.

```ts
  valueParts.forEach((_, i) => {
    if (meetings.length <= i) {
      // If there aren't enough meetings, create a new one
      meetings.push({
        days: [],
        duration: 0,
        location: { building: "", roomCapacity: 0, roomNumber: "" },
        startTime: "",
      });
    }
  });
```

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
```diff
@@ -27,12 +27,36 @@ const thursReg = RegExp("[Tt][Hh]|[Rr]");
...
+  return valueParts;
+};
+
+export const startTimeCallback = (value: string, params: CaseCallbackParams) => {
+  const startTimes = createMeetings(value, params);
+  const { meetings } = params;
+  startTimes.forEach((startTime, i) => {
+    meetings[i].startTime = startTimeCase(startTime);
+  });
```
charkour: I see this pattern of logic repeated often. Would it be better to make a function helper?

```ts
/* 
A function that creates meetings and assign them using a custom assignment function.
*/
const assignWithMeetings = (
  value: string,
  params: CaseCallbackParams,
  arrAssign: (value: string, i: number, arr: Meeting[]) => void,
) => {
  const valueParts = createMeetings(value, params);
  const { meetings } = params;
  valueParts.forEach((v, i) => {
    arrAssign(v, i, meetings);
  });
};
```

Use: 
```ts
export const startTimeCallback = (value: string, params: CaseCallbackParams) => {
  assignWithMeetings(value, params, (startTime, i, meetings) => {
    meetings[i].startTime = startTimeCase(startTime);
  });
};

export const locationCallback = (value: string, params: CaseCallbackParams) => {
  assignWithMeetings(value, params, (location, i, meetings) => {
    [meetings[i].location.building, meetings[i].location.roomNumber] = locationCase(location);
  });
};
```
Maybe this is better because it allows for this pattern to be reused easily. Let me know what you think.

charkour (changes requested):
This looks really good! Just two comments. The additions you made are awesome and I was able to export and import a CSV with seemingly matching data. I didn't try testing for the crash that Jon got earlier. Nice work.

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
Jdellis818: This is awesome!

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
bryantgeorge: 2453811

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
bryantgeorge: This is great, thanks: 2453811

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
bryantgeorge: 2453811 makes this less of an issue

charkour (approved):
Looks good to me! This is great work.

@Jonri2, I could not get the app to crash. It could handle (after some time) the Course Section Enrollment.csv. 

@Syobnaf, was this PR also supposed to fix this bug of creating two sections for the same professor? If not, then this PR is all set to go!
image

Jdellis818:
> @Jonri2, I could not get the app to crash. It could handle (after some time) the Course Section Enrollment.csv.

The crash was fixed after the first review iteration. I'm not sure what was causing it, but it seems good now :)

Jdellis818 (approved):
LGTM

PR #51, opened 2020-11-23
Fix duplicate events bug

### LINKED ISSUE #48: More than 1 room section when there are multiple professors

Jdellis818:
This should fix the duplicate events we were having. Check out Math-190 for Scofield on the math-schedule and you can see there's only 1 event now.


charkour (approved):
LGTM! I realized that for Bolt and the other profs, the overlaps because they teach two classes at the same time. Thanks for this fix.

PR #54, opened 2020-11-25
clean up repository

### LINKED ISSUE #53: Clean up repository

charkour:
This does not make any changes to the functionality of the program, but only to the code. This adds barreling and auto organized imports. 

Barreling is nice because it allows you to import many components from one folder. Check this out for more information [Barrelsby](https://github.com/bencoveney/barrelsby). I might consider adding this to automate the creation, but doing it by hand is not too difficult. 

To use barreling in the future, whenever you create a new folder, create an `index.ts` to export your component in that folder, and in the parent folder, export the name of the new folder you just created, and then you can use components relative to source (which is an option in `tsconfig.json`).

Having a clean repo is important for easy future development! Let me know what you think. Thanks.

Closes #53


charkour:
Hmm, it seems the re-arranging the imports after barreling has messed up the context. It works locally, but not on GitHub. I'll fix it in the morning.

charkour:
Okay, I figured out the bug. For some reason, you can only import contexts from `utilities/contexts` and not `utilities` in certain cases. If you remember that, then it's fine to use absolute imports without any concerns. I've disabled the ability to import context from `utilities` so ESLint will throw an error if you try that.

Jdellis818 (approved):
LGTM. Thanks for this work

PR #57, opened 2020-11-26
Organize popover menu form

### LINKED ISSUE #56: Organize Popover

charkour:
This is the first of a three-part PR involving the popover menu. In part one, the goal was the split up the logic and the structure of the popover component. Also, the goal of this was the remove potential mismatches of related data. For example, the component name depends on the component label, so this PR allows one to derive the name from the label. 

Let me know what you think! I tried to split up the code in a way I thought was logical, but if you disagree, let me know and we can find a better solution! 

Part 2 will be validating the form.
Part 3 will be testing the form.

Also, I didn't realize that Jon is editing the Popover so I apologize for the merge conflicts we will encounter at some point. 

closes #56

### DIFF
--- a/client-course-schedulizer/src/components/reuseables/AddSectionButton/AddSectionButton.tsx
+++ b/client-course-schedulizer/src/components/reuseables/AddSectionButton/AddSectionButton.tsx
@@ -2,11 +2,10 @@ import { Button, ButtonProps, IconButton, Popover } from "@material-ui/core";
 import { Add } from "@material-ui/icons";
 import { AddSectionPopover } from "components";
 import { bindPopover, bindTrigger, usePopupState } from "material-ui-popup-state/hooks";
-import React, { ComponentType } from "react";
+import React from "react";
 import "./AddSectionButton.scss";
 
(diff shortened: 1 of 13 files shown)


charkour:
I'd like to test this some more before it gets merged in.

charkour:
Okay, this is all set to be reviewed. There is currently a bug where starting a schedule from scratch causes the app to crash, but that'll be fixed in another PR.

bryantgeorge:
Wow, this looks a lot nicer! The organization looks good for now though it will likely need updating again when the additional fields needed for export are added (and when the location field is separated into building and room number). The organization of the code seems to make sense. There will likely be merge conflicts with my code as well (because I update the interfaces in #58).

bryantgeorge (approved):

charkour:
> Wow, this looks a lot nicer! The organization looks good for now though it will likely need updating again when the additional fields needed for export are added (and when the location field is separated into building and room number). The organization of the code seems to make sense. There will likely be merge conflicts with my code as well (because I update the interfaces in #58).

Thanks for the heads up! We'll make a PR in the future to resolve those new features.

PR #58, opened 2020-11-27
Improve Export

### LINKED ISSUE #39: Improve Export Function

bryantgeorge:
This branch aims to:
1. Eliminate bugs with commas in fields causing fields to be displaced (like if a course's name was `Proof, Logic, and Complexity`, then this would all go in the name field, rather than `Proof` going in the name field, and `Logic` and `and Complexity` going in the next two fields). Note that I don't try to prevent this with fields that obviously shouldn't have commas (like date fields), if you think this is a mistake, let me know and I will make all fields strict.
2. Make the export conform to the registrar's format (the bulk of the changes in this PR).
    - Note that the export should have all the same fields as the registrar's format, BUT it will also have some additional fields that are nice to have (and maybe should be incorporated into the registrar's format), these are: `Comments`, `SemesterLength`, and `MeetingDurationMinutes`.
    - I added fields to the data interfaces where necessary to be able to export everything, but these added fields are not currently fillable when using the add course/section popup (see relevant TODO).

For testing, I recommend importing the `Course Section Enrollment.csv` and `math-schedule.csv` and subsequently exporting them. It's really satisfying to see an output which is very similar to the input, but spot check some lines to see that the fields are matching up (some things will differ like the extra columns and online courses wont have any meeting info). Also make sure you can add a course/section without error.

(In review, if you ever feel that I leave too many TODOs rather than merely addressing them in this current PR, let me know and I will take a look)

Closes #39


client-course-schedulizer/src/utilities/helpers/writeCSV.ts
```diff
@@ -1,41 +1,115 @@
...
-        startTimeStr += `${meeting.startTime}\n`;
-        durationStr += `${meeting.duration}\n`;
-        locationStr += meeting.location.roomNumber
+        startMoment = moment(meeting.startTime, "h:mm A");
+        if (startMoment.isValid()) {
+          meetingTimeStr += `${startMoment.format("h:mmA")} - ${startMoment
+            .add(meeting.duration, "minutes")
+            .format("h:mm A")}\n`;
+          meetingStartStr += `${startMoment.format("h:mm A")}\n`;
+          meetingStartInternalStr += `${startMoment.format("H:mm:ss")}\n`;
```
Jdellis818: When you add to a moment, it reassigns the moment to the new value even if you don't explicitly reassign it. So, in this case, adding to `startMoment` with `.add()`when computing the end time results in `meetingStartStr` and `meetingStartInternalStr` equalling the end time rather than the start time. You'll have to make a copy of `startMoment` when you add to it.

Jdellis818 (changes requested):
This is really good. I'm glad you added more of the fields to import. The only issue I found was that the `MeetingStartTime` was incorrect when exporting. Otherwise, it looks good to me!

charkour (changes requested):
Thank you for adding tests!! This is great work.  The only issue I found was the same as Jon's. For example, Bolt's MATH-171-A class should start at 11:30 and end at 12:20, but it seems to be 10 minutes off. 
image


There's also a bug where if you start a schedule from scratch, the app crashes. We'll fix that in another PR. I'm trying to figure out what that is but don't really know why yet. Do you have any ideas? 

Ah, the starting the schedule from scratch bug relates to the function `getMinAndMaxTimes()`, but I am not sure what is causing it yet.

client-course-schedulizer/src/utilities/helpers/writeCSV.ts
bryantgeorge: Good catch: 1e28f58

bryantgeorge:
> Thank you for adding tests!! This is great work. The only issue I found was the same as Jon's. For example, Bolt's MATH-171-A class should start at 11:30 and end at 12:20, but it seems to be 10 minutes off.
> image
> 
> There's also a bug where if you start a schedule from scratch, the app crashes. We'll fix that in another PR. I'm trying to figure out what that is but don't really know why yet. Do you have any ideas?
> 
> Ah, the starting the schedule from scratch bug relates to the function `getMinAndMaxTimes()`, but I am not sure what is causing it yet.

I made b060070 to make the moment format more consistent, but I'm not sure if my changes have anything to do with the problem you observed. In fact, I wasn't able to reproduce this crash, what course did you try adding from scratch that led to the crash?

charkour:
>  In fact, I wasn't able to reproduce this crash, what course did you try adding from scratch that led to the crash?

I entered "2" for every field and picked random days the class met. I'll make an issue for the bug so someone can fix it in a later PR. 

Thanks for the great work!

charkour (approved):
LGTM

bryantgeorge:
Okay, the merge conflicts should be all fixed now.

charkour (approved):
Looks good! Thanks for the edits. The exported CSV looks good. It would be nice to get some tests added to make sure it still works if we refactor things in the future. 

The only issue I saw was if I edited a class on the schedule, the data got added incorrectly which I think is due to our editing function and than this exporting PR, right? For example, if I move Bolt's 171 class to 2:00am, then the data that gets exported is slightly different for this class than the rest of the classes.

client-course-schedulizer/src/utilities/helpers/readCSV.ts
```diff
@@ -95,7 +95,7 @@ export const csvStringToSchedule = (csvString: string): Schedule => {
     if (fields) {
       fields.forEach((field) => {
         const value = String(object[field]);
-        field = field.replace(/\s/g, "");
+        field = field.replace(/\s/g, "").replace("ï»¿", "");
```
charkour, line 98: haha what is that?

charkour:
Once the tests pass, I think this is good to merge (you can force merge it if you don't get all the reviews).

client-course-schedulizer/src/utilities/helpers/readCSV.ts
bryantgeorge: So I was importing my `Course Section Enrollment.csv`, and I noticed that the `Department` field wasn't being exported, and when I logged the field names in `readCSV`, the first one was `ï»¿Department`. (Note that `math-schedule.csv` did work). Apparently, some files have that at the beginning, but you can't see it in an editor: https://stackoverflow.com/questions/3255993/how-do-i-remove-%C3%AF-from-the-beginning-of-a-file

charkour: Wow, you learn something new each day. I didn't know that. Thanks!

PR #62, opened 2020-11-29
Make events clickable and editable

Jdellis818:
This feature allows the user to click on events on the calendar and edit them. To test, import a csv, click on an event, and change the values. When you click submit, the calendar should update correctly.

There are quite a few merge conflicts with Charlie's PR, so we'll have to figure those out before we can review.


client-course-schedulizer/src/components/Toolbar/ScheduleToolbar/ScheduleToolbar.scss
```diff
@@ -14,8 +14,8 @@
   padding: $spacing;
   margin-left: -2 * $spacing;
   margin-right: -2 * $spacing;
-  // TODO: there is slight squishing when stickied but not sure how to fix. takes time to debug
-  margin-bottom: 1.5 * $spacing;
+  // TODO: there is slight squishing when scrolling down before sticky headers get picked up
+  margin-bottom: 2.5 * $spacing;
```
charkour: this makes the header jerk on my mac. I know you were experiencing issues will the search bar being cut off, so maybe you could look into increasing the z-index on the search bar.

client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx
```diff
@@ -43,6 +46,20 @@ const ScheduleBase = ({ calendarHeaders, groupedEvents, ...calendarOptions }: Sc
...
+    variant: "popover",
+  });
+
+  const handleEventClick = useCallback(
+    (arg: EventClickArg) => {
+      setPopupData(arg.event.extendedProps as CourseSectionMeeting);
+      popupState.open(arg.el);
+    },
+    [popupState],
+  );
```
charkour, line 62: nice use of callback

client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx
```diff
@@ -79,14 +96,33 @@ const ScheduleBase = ({ calendarHeaders, groupedEvents, ...calendarOptions }: Sc
...
+                      events={filteredEvents[header]}
+                    />
                   </div>
                 );
               })}
             </div>
           </Stick>
         </div>
       </div>
+      <Popover
```
charkour, line 112: I think you'll want to make this a separate component and use it here and when you click the add button

charkour (changes requested):
Thanks for making some splitting up functions. It makes the code easier to read and test. Overall this is great work! Just a comment or two to look at before this is merged in.

Also, with the merge conflicts, if you'd prefer I can try to solve them. Just let me know.

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -61,35 +73,71 @@ export const AddSectionPopover = () => {
...
           register={register}
-          textFieldProps={{
-            defaultValue: "08:00",
-            fullWidth: true,
-            type: "time",
-          }}
+          textFieldProps={{ fullWidth: true, name: "startTime", type: "time" }}
+          value={
+            values?.meeting.startTime
+              ? moment(values?.meeting.startTime, "h:mma").format("HH:mm")
```
bryantgeorge, line 138: This will help to make things more consistent with the code in #58 
```suggestion
              ? moment(values?.meeting.startTime, "h:mm A").format("HH:mm")
```

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -48,6 +57,9 @@ export const AddSectionPopover = () => {
   const isIntensiveSemester = semesterLength === SemesterLengthOption.IntensiveSemester;
   const isCustomSemester = semesterLength === SemesterLengthOption.CustomSemester;
 
+  const locationValue =
+    (values && `${values?.meeting.location.building} ${values?.meeting.location.roomNumber}`) || "";
```
bryantgeorge, line 61: this results in the string `" "` when no `building` or `roomNumber` is specified (I think because `" "` returns true). I guess:
```ts
const locationValue = (values && (values.meeting.location.building || values.meeting.location.roomNumber)) ? `${values.meeting.location.building} ${values.meeting.location.roomNumber}` : ""
```
might work but looks clunky (maybe could also try using `trim()` with the current implementation)

bryantgeorge:
One minor possible improvement could be to make the class events cause the mouse to change to a hand icon when hovered over to make it clear that you can click on them (I don't expect this change to be made in this PR, just a thought for later).

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
charkour: Good catch, thanks.

bryantgeorge:
One final note is that I noticed that when adding/updating a section, its start time is exported in military time instead of `h:mm A` format, but this problem isn't specific to this branch because it happens in `develop` too, so I also don't necessarily expect it to be fixed here either.

bryantgeorge:
Merging now for demo, bugs will be addressed in separate issue

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
bryantgeorge: As a note, I think
```ts
const locationValue =
    ((values && `${values?.meeting.location.building} ${values?.meeting.location.roomNumber}`) || "").trim();
```
Might be the simplest/best approach

PR #67, opened 2020-12-02
Fix calendar label bugs

### LINKED ISSUE #61: App crashes when creating a schedule from scratch
charkour said this occurred when entering in `2` for all fields (with at least one day selected) and is related to `getMinAndMaxTimes()`. After some research, this seems to occur when the only class has a very short duration (like 2 minutes for example) because then the min and max times are very close to each other.

More specifically, if the only class in the schedule has a duration of 30 minutes or less, it will cause the app to crash (when I try a singular class with a duration of 31 minutes or greater it works).

### LINKED ISSUE #63: Time axis wrong if first class starts not on the hours (ie 2:30)

bryantgeorge:
When adding a class from scratch:
- Making the only class in the schedule have a duration of 30 minutes or below should no longer cause a crash
- The hour labels on the left of the schedule should be placed properly (especially when the course starts not on an hour)

Try to break this with strange course start times and durations. For testing, it is probably best to start fresh and add courses manually, and refresh the page to start fresh again every once and a while.

Closes #61 
Closes #63

### DIFF
--- a/client-course-schedulizer/src/utilities/services/scheduleService.ts
+++ b/client-course-schedulizer/src/utilities/services/scheduleService.ts
@@ -76,8 +76,11 @@ export const getMinAndMaxTimes = (schedule: Schedule) => {
   const endTimes = map(meetings, (meeting) => {
     return moment(meeting.startTime, "h:mma").add(meeting.duration, "minutes");
   });
-  const minTime = minBy(startTimes)?.format("HH:mm") || "6:00";
-  const maxTime = maxBy(endTimes)?.format("HH:mm") || "22:00";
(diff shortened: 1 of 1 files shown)


charkour (changes requested):
Thank you for fixing this! It works so much better now.

The only issue is that the calendar timing breaks if I add a class that starts at 11:00pm and goes for 120 minutes. It breaks if a class spans multiple days which I think is an edge case we shouldn't worry about. Would it be good for me to require users to not create classes that span multiple days on my `feature/popover-validation` branch?

charkour (approved):
I'll add the validation.

Issue #65, opened 2020-12-01
Change requested on #62


charkour:
Since we wanted to merge in #62 for the demo, here are changes requested that should be addressed.

* add `? moment(values?.meeting.startTime, "h:mm A").format("HH:mm")` there the `h:mm A` is the change when dealing with time. [link](https://github.com/senior-knights/course-schedulizer/pull/62#discussion_r532836599)
* [Fix the space given for location](https://github.com/senior-knights/course-schedulizer/pull/62#discussion_r532842834)
* [Turn the cusor to a hand when it is over an event](https://github.com/senior-knights/course-schedulizer/pull/62#issuecomment-735991218)
* ~~[generalize use of time](https://github.com/senior-knights/course-schedulizer/pull/62#issuecomment-736094772)~~ I believe this works properly in #58 
* Location and term is not respected when updating an event

Also, I suggested refactoring the popup button to another component, but I was thinking it might be better to make it a Higher Order Component to reuse and apply that logic. So that will be in a different PR.

bryantgeorge:
Note to self: update should always delete the section being updated?

PR #68, opened 2020-12-03
Feature/popover validation

### LINKED ISSUE #60: Validate Popover Form

charkour:
This feature adds form validation to the popover. It requires prefix, number, and section are added. This also makes use of a Form context provider which cuts down on repeated LOC. That's why the additions and deletions are high. 

To test:
- try to click submit on a pop over to see the errors and then try to fix them or pass in bad values. 
- test clicking a created event and then seeing that the data fill properly

Finally, I suggest we make the popover into a full-page in the future because there are so many options that need to be adjusted and we should have a simplified version of editing that is a popover. I'm basing a lot of the design decisions on Google Calendar.


client-course-schedulizer/src/utilities/services/validation/popover.ts
```diff
@@ -0,0 +1,63 @@
...
+  number: number().typeError("number must be a number").required().positive().integer(),
+  prefix: string().required().uppercase(),
+  roomCapacity: number().typeError("global max must be a number").positive().integer(),
+  section: string().required().uppercase(),
+  studentHours: number()
+    .typeError("student hours must be a number")
+    .required()
+    .positive()
+    .integer(),
+});
```
bryantgeorge, line 105: For the validation:
1. Having the error messages list fields in camel case is slightly strange, but not a big deal.
2. `facultyHours` and `studentHours` should be allowed to be decimals (like 1.5).
3. I don't think `location`, `days`, and `duration` should be required due to online asynchronous classes.
4. Similarly with `instructor`, I wonder if it shouldn't be required to allow for schedules where things aren't fully planned out yet?

client-course-schedulizer/src/utilities/services/addSectionService.ts
```diff
@@ -142,20 +203,23 @@ export const mapInputToInternalTypes = (data: SectionInput) => {
...
         },
         startTime: startTimeCase(data.startTime),
       },
     ],
     semesterLength: semesterType,
     studentHours: Number(data.studentHours),
     term: data.term,
-    year: "2021-2022",
+    // TODO: there currently isn't a way to get the year for this Section.
+    year: `${"hello"}`,
```
bryantgeorge: For testing with dept chairs soon, should we have it assume 2021? (Also I noticed in the `Course Section Enrollment.csv` that they list 2020 as the academic year for all classes, even if the class is in spring when it's 2021)

bryantgeorge:
I noticed that there's an `x` button on the time slot which allows you to clear the time (or by pressing backspace) which can lead to meetings without time, is this intentional? (This might be out of the scope of this PR because I assume this is how it always worked)

image

bryantgeorge (changes requested):
I like how convenient this is, good work! I agree that making it a separate page later will be good because of how big it is. Also I might have an idea for fixing the multiday classes bug, so I can look at that in #65.

client-course-schedulizer/src/utilities/services/validation/popover.ts
charkour: Yeah, those are good points. I will attempt to fix those.

charkour:
@Syobnaf, Thanks for the comments! I'll address these over the break since I don't think there is a rush to add validation.

charkour:
> I noticed that there's an `x` button on the time slot which allows you to clear the time (or by pressing backspace) which can lead to meetings without time, is this intentional? (This might be out of the scope of this PR because I assume this is how it always worked)
> 
> image

I couldn't reproduce this on Chrome MacOS. Maybe it is browser/OS specific?

*Edit*
I was able to reproduce it. I'll look into it but I think it'll be another PR

client-course-schedulizer/src/utilities/services/addSectionService.ts
charkour: I like your fix in pr #69 . Thanks

client-course-schedulizer/src/utilities/services/validation/popover.ts
```diff
@@ -0,0 +1,93 @@
...
+    .test("is-decimal", "invalid decimal", decimalRegex)
+    .transform(emptyStringToNull)
+    .nullable(),
+  year: number()
+    .typeError("year must be a number")
+    .positive()
+    .integer()
+    .transform(emptyStringToNull)
+    .nullable(),
+});
```
bryantgeorge, line 105: I think we may want to add checks for the `used` and `day10Used` fields like:
```ts
  used: number()
    .typeError("used must be a number")
    .positive()
    .integer()
    .transform(emptyStringToNull)
    .nullable(),
  day10Used: number()
    .typeError("day 10 used must be a number")
    .positive()
    .integer()
    .transform(emptyStringToNull)
    .nullable(),
```

client-course-schedulizer/src/utilities/services/validation/popover.ts
```diff
@@ -0,0 +1,93 @@
...
+  instructor: string(),
+  localMax: number()
+    .typeError("global max must be a number")
+    .positive()
+    .integer()
+    .transform(emptyStringToNull)
+    .nullable(),
+  location: string(),
+  name: string(),
+  number: number().typeError("number must be a number").required().positive().integer(),
```
bryantgeorge: Sorry I didn't consider this earlier, but should we allow a string here for classes like `SPAN-W80-A` where the number is a string like `W80`? See the full schedule for other examples like this.

bryantgeorge (changes requested):

client-course-schedulizer/src/utilities/services/validation/popover.ts
charkour: Ah, good catch! I forgot about that too. Thanks. We can just allow a string.

client-course-schedulizer/src/utilities/services/validation/popover.ts
charkour: Sounds good. Thanks!

bryantgeorge (approved):
LGTM

PR #69, opened 2020-12-09
Address popover issues

bryantgeorge:
This Branch:
1. Addresses issues in #65 
2. Makes the update popup different from the add popup (gives the option to either update the section, thus deleting the old version, or merely add a new section built off of the section clicked originally)
3. Adds in the missing fields to the popup, except for `startDate`, `endDate`, and `termStart`

To Test:
1. `npm i` and `npm start`
2. Import `math-schedule.csv` and try adding sections the regular way and playing around with updating sections (or adding sections from the update popup)
3. Ensure that updates in things like location, instructor, time, and term are reflected graphically (specifically notice what happens on updating MATH-171-A of Bolt to Spring)
3. Export and verify that changes are reflected

Notes:
1. This should be treated with harsh skepticism in testing because this could unintentionally introduce more issues than before, but in theory, it should make updating much more reliable.
2. This branch more closely associates sections with the instructor, such that MATH-100-A if taught by Bob Smith will be treated separately from MATH-100-A taught by Jane Doe in the same term (this doesn't mean that multiple teachers can't be assigned to a single section though). The reason for this change is that without doing so, changing the term of a class was adding it to another instructor's section of the same class (it's a bit confusing, so ask me if you want more details).
3. This will introduce conflicts with @charkour 's code in #68.
4. Let me know your thoughts on the differentiation of `Update Section` and `Add Section`, is this intuitive or just confusing?

Closes #68


bryantgeorge:
Okay, this should be good now (I fixed the full schedule import) ~~though I noticed that the time labels on the left still aren't totally aligned though I'm not sure why...~~ (fixed in 4d463fd, but maybe not the best solution?)

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -42,10 +42,17 @@ export const AddSectionPopover = ({ values }: AddSectionPopover) => {
     convertFromSemesterLength(values?.section.semesterLength).toLowerCase() as SemesterLengthOption,
   );
 
-  // handlers
-  const onSubmit = (data: SectionInput) => {
+  // Handlers
+  // TODO: Why doesn't it show up???
```
Jdellis818: This shows up now, right?

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -57,31 +64,76 @@ export const AddSectionPopover = ({ values }: AddSectionPopover) => {
...
+  const buttons = values ? (
+    <div>
+      <Button
+        className="update-button"
+        color="primary"
+        onClick={handleSubmit(onSubmitUpdate)}
+        variant="contained"
+      >
+        Update Section
+      </Button>
```
Jdellis818, line 81: I'm a big fan of adding this

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
```diff
@@ -150,6 +216,20 @@ export const AddSectionPopover = ({ values }: AddSectionPopover) => {
           }}
           value={(values?.meeting.duration || "").toString()}
         />
+        <GridItemTextField
+          label="Year"
+          register={register}
+          textFieldProps={{ name: "year" }}
+          value={(values?.section.year || "").toString()}
+        />
```
Jdellis818, line 223: In the long term, I think Year will be set for the whole schedule, but this is definitely fine for now

client-course-schedulizer/src/components/reuseables/Schedule/Schedule.scss
```diff
@@ -39,15 +43,14 @@ $header-offset: 8;
 .time-slot {
   display: flex;
   justify-content: flex-end;
-  margin-top: -1.5 * $spacing;
+  height: 50px; // TODO: Calculate or use constant?
```
Jdellis818, line 46: On my computer, there's something that is covering the bottom half of the click area for the buttons on the toolbar. I'm not sure if it is this or not

Jdellis818 (approved):
This looks great! It fixes almost all the issues with #65. The only issues I noticed was something related to styling which is making the bottom half of buttons unclickable and when the semester length is edited, if the edit popover is brough up again, the options to click "First" or "Second" is gone. With the first, I just checked and it's occurring on the prod brach as well so no need to address it here. With the second, it's a pretty small issue so I'm just  gonna add an issue for it and we can address it later. Great work!

client-course-schedulizer/src/components/reuseables/AddSectionPopover/AddSectionPopover.tsx
bryantgeorge: Oops, yes it does

client-course-schedulizer/src/components/reuseables/Schedule/Schedule.scss
bryantgeorge: #76

PR #70, opened 2020-12-10
Update issue templates

charkour:
In this PR:
- Two markdown files that are generated by Github

No need to test. 

This is for the future when we start to do support for stakeholders. These are the default issue templates from GitHub which I think give a really good starting point for creating issues. I've never used these before so please feel free to add suggestions.


bryantgeorge (approved):
LGTM, these are nice for standardizing feedback!

PR #77, opened 2021-01-03
Perf/optimize schedule headers

### LINKED ISSUE #59: Optimize Schedule for many-many calendars
Replace nested loops with maps (?).
Use memorization

charkour:
This PR adds some optimization to the views containing schedules (a long list of calendars). 

When uploading the `Course Section Enrollment.csv`, there were performance errors partially due to the sticky packages. Removing those improved the performance a good bit. I added the `react-window` package to only render ~35 calendars at a time which also improves performance. This also separates out the ScheduleHeader and LeftTimeAxis components. 

To Test: 
- Go to a commit before these changes and upload a large CSV (`git checkout 0fa2a69` for example). Test scrolling quickly and note the performance. It might be best to use a previously deployed version of the app. 
- Go to this branch and upload the same large CSV. Test scrolling quickly and note the smoother performance. 

A couple of notes:
- The built and deployed version will perform better than the development version.
- The calendar headers are no longer sticky. I didn't think it was needed, but we could add them back if we want to. It is somewhat tricky to do it in a performant manner.
- Scrolling in one direction and quickly scrolling in the other direction could cause the user to see blank calendars. If this is an issue, we can try using the `react-virtualized` package with has more control than the one I used.

Let me know what you think!

*Update*
I forgot to say this before, but `react-window` allows us to scroll the user to a specific element which will be a helpful feature in the future. 
Also, I think that the sticky header is nice so we can make an issue to add it back in the future, that'll just be hard. We might have to be creative and change how we are displaying the entire schedule.

closes #59


bryantgeorge (commented):
I tried this inside my VM on Chrome with the current `develop` page https://sharp-babbage-a45ee2.netlify.app/#/ (on b3c5dae) versus this branch (I ran `npm run build` and then `serve -s build`). Assuming I did this correctly, I didn't really notice that much of a performance difference between them (current `develop` actually seems really fast for me outside my VM). This branch has the added issue that scrolling left/right causes the columns to need to load in whereas current `develop` seems to keep them all loaded. Maybe I'll defer to @Jonri2 for further thoughts.

Code looks good though (so no need to re-review that Jon, just testing for performance is fine)

charkour:
@bryantgeorge, thanks for letting me know!

I agree this solution isn't perfect so I'll just turn this into a draft for now. The package I used is a technology I wanted to try. I still think we'll want to optimize the schedule in the future, but we can wait. I'd like to work on this later in the future, but I think we can get it to work better.

PR #83, opened 2021-02-07
Allow export of old and new formats

### LINKED ISSUE #80: Revise Output Formats
The format required by the registrar has changed to a simplified format for classes and non-teaching loads. We need the ability to export these formats.

See:
- Courses: [2021-2022 Course Schedule Template.xlsx](https://github.com/senior-knights/course-schedulizer/files/5928242/2021-2022.Course.Schedule.Template.xlsx)
- Non-Teaching: [2021-2022 Non-Teaching Load Template.xlsx](https://github.com/senior-knights/course-schedulizer/files/5928244/2021-2022.Non-Teaching.Load.Template.xlsx)

bryantgeorge:
To Test:
- `npm i && npm start`
- Try importing the full course schedule, the math schedule, etc.
- There should now be two export buttons, try them both:
  - `EXPORT CSV` should give two CSVs in the new formats described by VanderLinden in our most recent meeting (it doesn't have the extra headers, but I figure the CSV should be good enough and allow department chairs to just copy the contents over)
  - `EXPORT FULL CSV` should do what `EXPORT CSV` used to do

Worth noting that the non-teaching CSV given is composed of classes without Prefix and Number, unsure if this is what we want to do going forward. (Currently it is impossible to add such a "course", so eventually we will want an add non-teaching load option, but it might still be good to store this as a course/section without prefix and number).

Let me know if the wording is confusing, if we should just remove full export or not, and if we should wait to merge this until we know more about the new formats.

Additionally, this branch changes the test CSV files to be loaded from their current version rather than current `develop` (I should have done it that way in the first place)

Closes #80


client-course-schedulizer/package.json
```diff
@@ -56,14 +56,15 @@
     "@types/react-stickynode": "3.0.1",
     "@types/react-table": "^7.0.25",
     "@types/yup": "^0.29.9",
-    "isomorphic-fetch": "^3.0.0",
+    "fs": "0.0.1-security",
```
charkour: No need for a change, just wondering what the reason is for a package change here?

bryantgeorge: This was due to changing from pulling the test CSV files from github via a url (with isomorphic-fetch) to grabbing them locally using fs and another package which was already installed. I guess I'm not sure if isomorphic-fetch could have been used to do the same thing or not, but the first solution I found didn't.

charkour: Cool! Thanks for finding a solution.

client-course-schedulizer/src/components/Tabs/tabComponents/CSVActions/FullExportButton/FullExportButton.tsx
```diff
@@ -0,0 +1,16 @@
...
+*/
+export const FullExportButton = (btnProps: ButtonProps) => {
+  const onFullExportClick = useExportFullCSV();
+
+  return (
+    <Button color="secondary" component="span" onClick={onFullExportClick} {...btnProps}>
+      Export Full CSV
+    </Button>
+  );
+};
```
charkour, line 16: This file, along with the other export/import buttons aren't used anywhere now. But I think it'll be nice to have if we change up the interface to use buttons rather than menu items.

client-course-schedulizer/src/components/Tabs/tabComponents/CSVActions/ExportButton/ExportButton.tsx
```diff
@@ -2,7 +2,7 @@ import { Button, ButtonProps } from "@material-ui/core";
 import React from "react";
 import { useExportCSV } from "utilities";
 
-/* A button that will export the current schedule as a CSV.
+/* A button that will export the current schedule as a CSV with the current simplified format.
```
charkour: What is the "current specified format". Is that one was specified by the registrar? The current statement seems ambiguous.

client-course-schedulizer/src/utilities/helpers/writeCSV.ts
```diff
@@ -1,114 +1,88 @@
...
+        // Create strings for fields that need to be constructed
+        const termStr = `${
+          typeof section.year === "number"
+            ? section.term === Term.Fall
+              ? String(section.year).slice(-2)
+              : String(section.year + 1).slice(-2)
+            : section.term === Term.Fall
+            ? String(Number(section.year.slice(-2)) - 1)
+            : section.year.slice(-2)
+        }/${section.term}`;
```
charkour: This block of code is duplicated, would probably be a good candidate for its own function.

client-course-schedulizer/src/utilities/helpers/writeFullCSV.ts
```diff
@@ -0,0 +1,115 @@
...
+      let monStr = "";
+      let tuesStr = "";
+      let wedStr = "";
+      let thursStr = "";
+      let friStr = "";
+      section.meetings.forEach((meeting) => {
+        startMoment = moment(meeting.startTime, "h:mm A");
+        endMoment = startMoment.clone().add(meeting.duration, "minutes");
+        if (startMoment.isValid()) {
+          meetingTimeStr += `${startMoment.format("h:mmA")} - ${endMoment.format("h:mmA")}\n`;
```
charkour: Similarly, this is also repeated in multiple files. Would it be bad if these were functions? I don't think it would hurt flexibility and then we would only have to make a change in one spot.

charkour (changes requested):
The code works good but maybe you could change the wording to be `Export CSVs` or `Export Registrar Format`. Does the registrar schedule cut too much info? For example, if we decide to allow the registrar format to be uploaded, then is there enough data to re-create the schedule? Currently, uploading the simple schedule doesn't work. But if we add that feature it would be another PR.

Great work overall! The code is good for the most part and it is cool that we can support different export methods.

client-course-schedulizer/src/components/Tabs/tabComponents/CSVActions/ExportButton/ExportButton.tsx
bryantgeorge: e0be0f3

client-course-schedulizer/src/utilities/helpers/writeCSV.ts
bryantgeorge: e0be0f3 (leaving duplicated code is probably one of my biggest flaws as a developer)

client-course-schedulizer/src/utilities/helpers/writeFullCSV.ts
bryantgeorge: e0be0f3

client-course-schedulizer/src/utilities/helpers/writeCSV.ts
charkour: If that's your biggest flaw, then you're a great developer! Keep up the quality work :)

bryantgeorge: \* one of, I've certainly got others

charkour (approved):
Boom. Looks good. If you want to update the wording for the Export Button you can, or we can take a look at it later after user feedback. Thanks!

Jdellis818:
> Boom. Looks good. If you want to update the wording for the Export Button you can, or we can take a look at it later after user feedback. Thanks!

What do you guys think about "Export Draft CSV" for the full version and "Export Final CSV" for the registrar version? That way, the user would save as much data as possible when simply drafting, but would export in the correct registrar format when ready to submit.

bryantgeorge:
Oops, I forgot to change the wording, I'll do that later today, thanks!

PR #85, opened 2021-02-08
Allow import from template spreadsheet

Jdellis818:
This feature makes the new spreadsheet template that the Registrar provided compatible to import into the Schedulizer. To test, save the template as a CSV and import it into the Schedulizer. You should see 1 course taught by John Calvin.


charkour (approved):
Works great! I believe there will be another PR that adds the ability to read the teaching loads CSV right? Maybe it would be good to add the template CSV to the repo.

PR #86, opened 2021-02-08
Chore/organize package json

charkour:
This is just a chore to move some of the type dependencies to the dev dependencies. This functionally doesn't change anything but it correctly labels the types as dev deps. While programming, we need to know what the types our, but our built JS app does not need to know types.


Jdellis818 (approved):
LGTM. 21,000 additions in the package-lock is crazy though

charkour:
> LGTM. 21,000 additions in the package-lock is crazy though

Haha yeah not sure why npm did that.

PR #94, opened 2021-02-08
Delete sections

### LINKED ISSUE #89: Delete Section/Course
There is no way to delete courses/sections from the schedule.

Have an option to delete the section (?) when clicking it on the schedule.

bryantgeorge:
This should work basically as you would expect:
- Try making some courses and/or importing a schedule
- Click on one of them in the calendar, the edit course/section popup should now have a `Delete Section` button, ensure that pressing it deletes the section
- Make sure to try this with a section of a course that has multiple sections (to make sure it doesn't delete all sections of the course)
- Make sure that the popup for adding a new course/section doesn't have the delete button (because you can't delete that which doesn't exist)

Closes #89


charkour (approved):
Looks great! Works well. Nice use of hooks.

PR #100, opened 2021-02-09
Use return to submit add section form

### LINKED ISSUE #96: "return to submit" isn't working
This doesn't work for me in Chrome on a Mac.

image

bryantgeorge:
I had removed this feature a while back (98db0ed) because I thought it might cause problems for users that happen to press enter/return without knowing it would submit the popup, but apparently we kept the note at the bottom saying that return submits, and Pruim seems to want it as a feature.

To Test:
- `npm i && npm start`
- Fill out the add a section popup, and press enter/return, it should correctly add your section.
- Try clicking that section to bring up the edit a section popup, change something and press submit. Your section should be updated (note that pressing enter triggers the update section button **not** the add a section button).
- I fixed a bug where adding a section with the same prefix, number, letter, term, and instructors would add that section as a new section rather than a new meeting on the existing section. To test this, try opening the update section popup and change just the time of the section, then press add section, and finally export. That section should have two meetings and should not be listed as two separate sections.

I think the huge deletion from `package-lock.json` is because I ran `npm i` and something got updated?

Closes #96


charkour (approved):
At first, it seemed bad to have two buttons with `type="submit"` but I think it works as intended as you said. 

Thanks for the bug fix as well.

PR #105, opened 2021-02-11
Fix Import Button Hitbox

### LINKED ISSUE #76: Bottom half of toolbar buttons unclickable
See https://github.com/senior-knights/course-schedulizer/pull/69#discussion_r546119240

bryantgeorge:
To Test:
- Make sure that the import button when initially loading the page can be clicked anywhere:

image

- Make sure that the other import button can also be clicked anywhere (especially on the bottom, top, and sides):

image

Closes #76


charkour (approved):
Nice! It works great. I'm impressed you found such a sneaky bug like that. And you just had to switch the order of two components. Cool beans.

PR #107, opened 2021-02-14
read xlsx schedules

### LINKED ISSUE #87: Add Excel Document Support for import.
I'm always frustrated when I have to convert the `.xlsx` to `.csv`.

Ability to update a `.xlsx` file. 

**Describe alternatives you've considered**
I have considered using a program to convert `.xlsx` to `.csv`.

**Additional context**
Recommend using `xlsx` package on npm.

charkour:
Adds the ability to upload `.xlsx` (Excel) files to the course schedulizer. 

To Test
- `npm i && npm start`
- Try uploading the .xlsx math schedule from `csv/`
- Next (perhaps in a new tab), upload the .csv version of the math schedule and notice the same schedule is uploaded.

I am proposing this change since I will like it will make the app easier to use for department chairs who might have trouble converting from one file format to another. 

The code is a little ugly, so I would be open to suggestions to make it look nicer. 

closes #87


charkour:
Not sure why my machine keeps adding more things to the `package-lock.json`

Jdellis818 (changes requested):
I definitely think this is a necessary feature. Great work on getting this implemented! I just have a few things to comment on before approving. First, we should change the terminology of our import buttons from "CSV" to "Spreadsheet". Second, while it does successfully import the math-schedule, it does not successfully import the Excel spreadsheet that the department chairs would get because it has the department name in the first row rather than the headers. It's easy enough to delete the title row in the spreadsheet, but department chairs may not know that they need to do that so maybe there's some way for us to check if the import failed and then try again by deleting the first row? I'm not sure exactly, but that's just a suggestion. I had a crude fix for a similar issue when importing the template format by looking to see if the CSV started with a 2, and deleting the first two rows if it did (since the template format started with "2021-2022") but I think it'd be much better to get a more generalized fix.

charkour:
@Jonri2, thanks for the review! I've made the naming change and I totally overlooked the new format for the schedule. I'll make a new issue for that since I don't think that is in the scope of this issue. I've made the issue #109 to track it.

Jdellis818 (approved):
LGTM!

PR #108, opened 2021-02-16
Fix semester length radio buttons

### LINKED ISSUE #75: Intensive and half semester radio options do not show up when passed data
To replicate, click on an Interim class. When it comes up, there are no options to change the intensive semester from "A" to something else.

bryantgeorge:
To Test:
- Upload the math schedule CSV, go to the interim, and click on a class.
- The update section popup should show that the section is an intensive, and it should show the intensive letter A. (Previously there was a bug which prevented the intensive letter radio buttons from displaying).
- Now try updating the section to be a different semester length (like a second half class or an intensive C class). When you change the `Semester Length` from `Intensive` to `Half`, `First` should be set by default (rather than nothing being selected).
- Click on that section again to display the update section popup again. Your change should have persisted. (There was a bug previously which caused the changes to not persist). Try exporting as well to ensure that the change is reflected in the CSV.

Closes #75


Jdellis818 (approved):
LGTM. Great work!

PR #115, opened 2021-02-19
save app state between sessions

### LINKED ISSUE #99: Save working schedule in localStorage and retrieve it.
I'm always frustrated when I navigate away and lose the schedule progresses. 

The current schedule state is saved in `localStorage`

**Describe alternatives you've considered**
I have tried to be careful not to leave the page while editing and have also periodically downloaded schedules. 

**Additional context**
I would recommend using [this package](https://github.com/streamich/react-use/blob/master/docs/useIdle.md) to periodically save to `localStorage` and also save when the user navigates away from the page.

charkour:
This saves the app state between sessions in localStorage. This means that it will only save the app state for the current web browser. This allows the user to leave the web page and come back and continue editing where they left off. 

To test:
- `npm start`
- Upload a schedule
- Refresh the page or navigate backward and notice the schedule loads when you return
- Change the semester and refresh again; notice the state persists through app sessions

This PR:
- a `useLocal` hook to help with modifying values in local storage
-   helpers for the hook
- updates to `initalAppState`
- changes to the main `reducer` function to be a closure that will accept a callback function when initialized

closes #99 

Please let me know any comments or changes!


Jdellis818 (approved):
This is awesome! It works exactly how I think users would want where refreshing or going back then forward again loads the schedule, but closing the tab and reopening a new tab goes back to the main unloaded screen. Nice work! The code is really slick too

charkour:
@Jonri2, thanks for the review. What do you mean by "closing the tab and reopening a new tab goes back to the main unloaded screen", specifically, what is the main unloaded screen?

Jdellis818:
Hmmm... For some reason, the first time I tested it, when I closed out the tab and reopened it, it did not load anything from localStorage and just functioned as it did before. But now it's loading in when I close and reopen the tab so I don't really know what happened that first time.

Jdellis818:
Yeah, I'd say just disregard that comment since I can't reproduce it. I may have accidentally opened a different version of the schedulizer when I was testing

charkour:
Okay thanks for checking. Right now it should load the schedule when you close and reopen the tab. We can change that in the future if it's weird.

PR #116, opened 2021-02-21
Color by feature

bryantgeorge:
Make sure to `npm i`. This branch should work about how you would expect it to, but classes with multiple prefixes or instructors won't be multicolored, so `"Bob Smith, Jane Doe"`, `"Bob Smith"`, and `"Jane Doe"` would all be colored uniquely. I don't think full calendar will allow multiple colors (please correct me if I'm mistaken). Is this solution optimal/intuitive in these cases?

Also in testing, make sure there isn't an overly noticeable slowdown when choosing color by or changing the schedule (if there is too much slowdown, maybe we eventually want to avoid coloring everything on each change).


charkour:
I know this PR isn't ready for review, but I couldn't wait to check this out! This is SO COOL. Using the deployed preview, I didn't notice any slowdown for changing the colors. Also, I think having `"Bob Smith, Jane Doe"`, `"Bob Smith"`, and `"Jane Doe"` be different colors works nicely. 

I look forward to reviewing this once it is set and good to go.

charkour (approved):
This looks great! Nice use of context and packages that already solved the problem. This feature is awesome. Ready to merge.

PR #118, opened 2021-03-02
Import csv by url

### LINKED ISSUE #92: Import CSV through URL

bryantgeorge:
To Test:
- Visit the app, but append something like `?xlsx=https://raw.githubusercontent.com/senior-knights/course-schedulizer/develop/client-course-schedulizer/csv/math-schedule.xlsx` or `?csv=https://raw.githubusercontent.com/senior-knights/course-schedulizer/develop/client-course-schedulizer/csv/math-schedule-full.csv` to the end of the url.
- It should load the schedule from that url.

Notes:
- 404s will be displayed in the console.
- I didn't worry too much about invalid files, but I think we can expect users to provide valid files (or notice their own error when not)? It seems that providing a csv when you say it's an xlsx or vice versa might actually still work based on my testing.
- The file in the url will only load if the initial state has an empty schedule AND the file should only load once on page refresh (try importing a different schedule after page load, it should work as expected). This is to prevent user work from being overwritten, but I can see how it might be confusing that the file provided in the url doesn't load if you have an existing schedule loaded, thoughts on this are appreciated.


client-course-schedulizer/src/components/App/App.tsx
```diff
@@ -13,6 +21,68 @@ export const App = () => {
...
+  useEffect(() => {
+    const scheduleHasCourses = schedule.courses.length > 0;
+    if (!scheduleHasCourses) {
+      // eslint-disable-next-line no-restricted-globals
+      const url = location.href;
+      // TODO: We'll need to be more careful about this if we end up using more GET params
+      const csvIndex = url.indexOf("?csv=");
+      const xlsxIndex = url.indexOf("?xlsx=");
+      if (csvIndex !== -1) {
+        const csvUrl = url.slice(csvIndex + 5);
```
charkour: What is this number's (5) importance? I would suggest making it a constant

client-course-schedulizer/src/components/App/App.tsx
```diff
@@ -13,6 +21,68 @@ export const App = () => {
...
+              if (result) {
+                const newSchedule = csvStringToSchedule(result);
+                await updateScheduleInContext(schedule, newSchedule, appDispatch, setIsCSVLoading);
+              } else {
+                setIsCSVLoading(false);
+              }
+            });
+        }
+      } else if (xlsxIndex !== -1) {
+        const xlsxUrl = url.slice(xlsxIndex + 6);
```
charkour: Again, what is the context of this number 6?

client-course-schedulizer/src/components/App/App.tsx
```diff
@@ -13,6 +21,68 @@ export const App = () => {
...
+              return response.arrayBuffer();
+            })
+            .then(async (result) => {
+              await appDispatch({ payload: { fileUrl: xlsxUrl }, type: "setFileUrl" });
+              if (result) {
+                const uploadedData = new Uint8Array(result as ArrayBufferLike);
+                const workBook = read(uploadedData, { type: "array" });
+                const firstSheet = workBook.Sheets[workBook.SheetNames[0]];
+                const newSchedule = csvStringToSchedule(utils.sheet_to_csv(firstSheet));
+                await updateScheduleInContext(schedule, newSchedule, appDispatch, setIsCSVLoading);
```
charkour: I think this should be a function since it is used in this file and `useImportFile.ts`. In both cases, it is serving the same function and changes to one would always be changes to the other.

client-course-schedulizer/src/components/App/App.tsx
```diff
@@ -13,6 +21,68 @@ export const App = () => {
...
+          setIsCSVLoading(true);
+          fetch(csvUrl)
+            .then((response) => {
+              if (response.status === 404) {
+                return null;
+              }
+              return response.text();
+            })
+            .then(async (result) => {
+              await appDispatch({ payload: { fileUrl: csvUrl }, type: "setFileUrl" });
```
charkour: On line 40, we check against the `fileUrl` to make sure we aren't loading the schedule repeatedly if we have already loaded it once, right? That is why we set the `fileUrl` after checking `csvUrl !== fileUrl`?

client-course-schedulizer/src/components/App/App.tsx
```diff
@@ -13,6 +21,68 @@ export const App = () => {
   const [isCSVLoading, setIsCSVLoading] = useState(false);
   const { save } = useLocal("appState");
   const [appState, appDispatch] = useReducer(reducer(save), initialAppState);
+  const { schedule, fileUrl } = appState;
+
+  // Load a file from the GET parameters (either ?csv= or ?xlsx=),
+  // only do this once for each page refresh and only do this if the initial state is empty
+  // Ref: https://stackoverflow.com/questions/56663785/invalid-hook-call-hooks-can-only-be-called-inside-of-the-body-of-a-function-com
+  // Ref: https://www.gitmemory.com/issue/SheetJS/js-xlsx/1110/531003740
+  useEffect(() => {
```
charkour: This effect is good, but I would move this effect into its own hook, like `useScheduleFromUrl()` and then call that in `App.tsx`. This separates the logic of getting the schedule from the app setup.

client-course-schedulizer/src/utilities/hooks/useLocal.ts
```diff
@@ -7,7 +7,12 @@
  */
 export const loadLocal = <T>(key: string) => {
   const item: string | null = localStorage.getItem(key);
-  return item ? (JSON.parse(item) as T) : undefined;
+  const thing = item ? JSON.parse(item) : undefined;
+  // Clear fileUrl
+  if (thing !== undefined) {
+    thing.fileUrl = "";
+  }
+  return thing ? (thing as T) : undefined;
```
charkour: This is good, except that `loadLocal` is a generic function that can't guarantee T will have the property `fileUrl`. This error can be seen by typing `thing` as `T | undefined` --> `const thing: T | undefined = item ? JSON.parse(item) : undefined;`. `loadLocal` is good for loading values from localStorage but assigning values to the object from storage is best done in another function. For example, create a function similar to the code below, which would be used in `appInterfaces.ts` in place of `loadLocal("appState")` 


```tsx
const loadInitalState = (): AppState | undefined => {
  const savedState = loadLocal<AppState>("appState");
  if (savedState !== undefined) {
    savedState.fileUrl = "";
  }
  return savedState;
}
```

charkour (changes requested):
This is some great work! I love this ability to import a schedule using a new method. This will really help when department chairs want to share a schedule. I like the decisions you made with the direction of this implementation and adding `fileUrl` to `appState` is a great idea. It works well with `useEffect()`. I think there are a couple things to address before we merge this feature in, but this is a great start!

In addition to other review comments, I would suggest that when a user types in `<base-url>/#/?csv=https://raw.githubusercontent.com/senior-knights/course-schedulizer/develop/client-course-schedulizer/csv/math-schedule-full.csv` it should override the current work and then redirect to `<base-url>/#/` and we can display on the app the `fileUrl` so people remember what it is. I think it is a little confusing that I can add a file to the URL but it won't actually display (if I have previous work). 

Maybe in the future, we can add a model that says "Importing a schedule from a URL will overwrite previous work, are you sure you want to continue?" to prevent unintended overwrites. Also, in the future it would be good for the Import Schedule button to allow a user to pick a file or paste in a URL since adding a file path to the URL can be difficult for some users. 

These are just my thoughts and I may be wrong. Great work! This feature is fantastic. No rush on the changes.

client-course-schedulizer/src/components/App/App.tsx
bryantgeorge: Oh, I guess I should make `"?csv="` a variable and use `.length`

client-course-schedulizer/src/components/App/App.tsx
bryantgeorge: Yep, otherwise every time the app is loaded, it loads the url's file (though maybe this problem could be solved if we redirect once the csv is loaded...)

client-course-schedulizer/src/components/App/App.tsx
bryantgeorge: 686213d but I called it in `SchedulizerPage.tsx`

client-course-schedulizer/src/components/App/App.tsx
bryantgeorge: 893fc3c

client-course-schedulizer/src/components/App/App.tsx
bryantgeorge: 893fc3c

client-course-schedulizer/src/components/App/App.tsx
bryantgeorge: 8256a59 but I didn't include the last two lines exactly because it makes the function more general (and usable in 3 places)

charkour:
Maybe the redirect isn't a big deal. So you can probably skip it if you haven't implemented it yet.

bryantgeorge:
> Maybe the redirect isn't a big deal. So you can probably skip it if you haven't implemented it yet.

Okay, although I was thinking of adding a button to import by url, so I may still end up adding this here.

client-course-schedulizer/src/components/App/App.tsx
charkour: Okay, awesome!

client-course-schedulizer/src/utilities/hooks/useLocal.ts
bryantgeorge: See c8c0a29, I didn't end up using this because it's actually better to not clear the file url

bryantgeorge:
c8c0a29 will reload without the search parameters in the url, and it displays the url which was loaded (let me know if the styling of this should be changed)

client-course-schedulizer/src/components/pages/SchedulizerPage/SchedulizerPage.tsx
```diff
@@ -1,7 +1,12 @@
 import { Tabs } from "components";
-import React from "react";
+import React, { useEffect } from "react";
+import { useImportRemoteFile } from "utilities/hooks/importRemoteFileHook";
 
 export const SchedulizerPage = () => {
+  // Load a remote file from the GET parameters
+  const { importRemoteFile } = useImportRemoteFile();
+  useEffect(importRemoteFile);
```
charkour, line 8: Beautiful and simple. Much more readable. It explains what the component does on each render without showing all the details.

client-course-schedulizer/src/utilities/hooks/useLocal.ts
```diff
@@ -7,7 +7,8 @@
  */
 export const loadLocal = <T>(key: string) => {
   const item: string | null = localStorage.getItem(key);
-  return item ? (JSON.parse(item) as T) : undefined;
+  const thing = item ? JSON.parse(item) : undefined;
+  return thing ? (thing as T) : undefined;
```
charkour: Is there a reason why there is the intermediary step with `thing`?

```suggestion
  return item ? (JSON.parse(item) as T) : undefined;
```

charkour (changes requested):

charkour:
Once that comment is addressed, this is good! Thanks for the great work.

client-course-schedulizer/src/utilities/hooks/useLocal.ts
bryantgeorge: Oops, I was trying to return it to how it was before, but I forgot about that, thanks for the suggestion!

PR #120, opened 2021-03-02
Make table cells clickable

Jdellis818:
This PR allows the user to edit sections from the faculty load view. To test, click on a section name in the faculty loads table. This will bring up the section edit popover with a paginator at the top. Use the paginator to find the section you are trying to edit, and make the change. This feature prevents the need for the user to switch to the faculty schedule tab and scroll to find the professor after seeing an issue in the faculty loads tab.

PR #130, opened 2021-03-04
Edit Other Duties

Jdellis818:
This PR allows the user to edit other duties on the teaching loads tab. To test, add a non-teaching activity. Then click on that activity in the table and change one of the fields. It should update successfully. Deleting a non-teaching load should work as well.

### DIFF
--- a/client-course-schedulizer/src/components/Tabs/FacultyLoads/UpdateNonTeachingLoadModalPagination/UpdateNonTeachingLoadModalPagination.tsx
+++ b/client-course-schedulizer/src/components/Tabs/FacultyLoads/UpdateNonTeachingLoadModalPagination/UpdateNonTeachingLoadModalPagination.tsx
@@ -4,6 +4,11 @@ import React, { forwardRef, Ref } from "react";
 import { findNonTeachingLoad, UpdateModalPaginationRef } from "utilities";
 import "./UpdateNonTeachingLoadModalPagination.scss";
 
+/**
+ * Creates a popup modal used on the FacultyLoads page to find
+ *   non-teaching loads and allow the user to page through and edit them
+ *   on a modal.
+ */
 export const UpdateNonTeachingLoadModalPagination = forwardRef(
   (props, ref: Ref<UpdateModalPaginationRef>) => {
     return (
--- a/client-course-schedulizer/src/components/Tabs/FacultyLoads/UpdateNonTeachingLoadModalPagination/UpdateNonTeachingLoadModalPagination.tsx
+++ b/client-course-schedulizer/src/components/Tabs/FacultyLoads/UpdateNonTeachingLoadModalPagination/UpdateNonTeachingLoadModalPagination.tsx
@@ -1,93 +1,19 @@
 import { ModalPagination } from "components";
 import { AddNonTeachingLoadPopover } from "components/reuseables";
-import React, {
-  ChangeEvent,
-  forwardRef,
-  Ref,
-  useContext,
-  useImperativeHandle,
-  useState,
-} from "react";
-import {
-  CourseSectionMeeting,
-  CourseSectionMeetingInstructorNonTeachingLoads,
-  findNonTeachingLoad,
-  Instructor,
-  UpdateNonTeachingLoadModalPaginationRef,
-} from "utilities";
-import { AppContext } from "utilities/contexts";
+import React, { forwardRef, Ref } from "react";
+import { findNonTeachingLoad, UpdateModalPaginationRef } from "utilities";
 import "./UpdateNonTeachingLoadModalPagination.scss";
 
-interface InstructorNonTeachingLoads {
-  instructor: Instructor;
-  nonTeachingLoads: string[];
-}
-
 export const UpdateNonTeachingLoadModalPagination = forwardRef(
-  (props, ref: Ref<UpdateNonTeachingLoadModalPaginationRef>) => {
-    const {
-      appState: { schedule },
-    } = useContext(AppContext);
-
-    const [open, setOpen] = useState(false);
-    const [modalValues, setModalValues] = useState<CourseSectionMeeting>();
-    const [instructorNonTeachingLoads, setInstructorNonTeachingLoads] = useState<
-      InstructorNonTeachingLoads
-    >({
-      instructor: "",
-      nonTeachingLoads: [],
-    });
-    const [page, setPage] = useState(1);
-    const handlePageChange = (event: ChangeEvent<unknown>, value: number) => {
-      setPage(value);
-      const courseSectionMeeting = findNonTeachingLoad(
-        schedule,
-        instructorNonTeachingLoads.nonTeachingLoads[value - 1],
-        instructorNonTeachingLoads.instructor,
-      );
-      if (courseSectionMeeting !== null) {
-        setOpen(false);
-        setModalValues(courseSectionMeeting);
-        setOpen(true);
-      }
-    };
-
-    const handleModalClose = () => {
-      setPage(1);
-      setOpen(false);
-    };
-
-    const handleModalOpen = ({
-      csm,
-      nonTeachingLoads,
-      instructor,
-    }: CourseSectionMeetingInstructorNonTeachingLoads) => {
-      if (csm) {
-        setModalValues(csm);
-        setInstructorNonTeachingLoads({
-          instructor,
-          nonTeachingLoads,
-        });
-        setOpen(true);
-      }
-    };
-
-    useImperativeHandle(ref, () => {
-      return {
-        handleModalOpen,
-      };
-    });
-
+  (props, ref: Ref<UpdateModalPaginationRef>) => {
     return (
       <ModalPagination
-        onClose={handleModalClose}
-        onPageChange={handlePageChange}
-        open={open}
-        page={page}
-        pageCount={instructorNonTeachingLoads.nonTeachingLoads.length}
-      >
-        <AddNonTeachingLoadPopover values={modalValues} />
-      </ModalPagination>
+        ref={ref}
+        findCourseSectionMeeting={findNonTeachingLoad}
+        PopoverComponent={AddNonTeachingLoadPopover}
+      />
     );
   },
 );
+
+UpdateNonTeachingLoadModalPagination.displayName = "UpdateNonTeachingLoadModalPagination";
--- a/client-course-schedulizer/src/components/Tabs/FacultyLoads/UpdateNonTeachingLoadModalPagination/UpdateNonTeachingLoadModalPagination.tsx
+++ b/client-course-schedulizer/src/components/Tabs/FacultyLoads/UpdateNonTeachingLoadModalPagination/UpdateNonTeachingLoadModalPagination.tsx
@@ -1,0 +1,93 @@
+import { ModalPagination } from "components";
+import { AddNonTeachingLoadPopover } from "components/reuseables";
+import React, {
+  ChangeEvent,
+  forwardRef,
+  Ref,
+  useContext,
+  useImperativeHandle,
+  useState,
+} from "react";
+import {
+  CourseSectionMeeting,
+  CourseSectionMeetingInstructorNonTeachingLoads,
+  findNonTeachingLoad,
+  Instructor,
+  UpdateNonTeachingLoadModalPaginationRef,
+} from "utilities";
+import { AppContext } from "utilities/contexts";
+import "./UpdateNonTeachingLoadModalPagination.scss";
+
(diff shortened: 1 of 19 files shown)


bryantgeorge:
I think this branch contains all of the changes from #120 ? Should I review this, and approve both?

bryantgeorge:
It seems that you can only click the non-teaching loads to edit if the teacher also has a regular course (otherwise clicking the non-teaching load does nothing)

bryantgeorge:
This branch seems to cause blank values to be filled in with 0s. To replicate, make a course with a minimal number of fields filled in (prefix, number, section letter, room, days, duration, and instructor), then click on the section to edit, and it will have 0s in many of the previously blank fields, like so:

image

bryantgeorge:
I don't understand what makes some items have a number beside them while others don't:

image

Is this intentional?

bryantgeorge:
Might be nice to switch the mouse cursor to the pointing hand icon when hovering over a clickable non-teaching load or section.

bryantgeorge:
If multiple semesters are selected, should that multiply the value of faculty hours? (This might be another judgement call)

client-course-schedulizer/src/utilities/services/validation/popover.ts
```diff
@@ -33,41 +33,36 @@ NOTE: fields with default values are not check: semester and time. */
...
   duration: number()
     .typeError("duration must be a number")
-    .positive()
     .integer()
     .transform(emptyStringToNull)
     .nullable(),
   facultyHours: number()
     .typeError("faculty hours must be a number")
-    .positive()
+    .integer()
```
bryantgeorge: I think we don't want the integer check here. Also, I think we want to leave the positive checks? I've been able to update courses to have a negative global max for example which seems illogical.

bryantgeorge:
Sometimes I'll edit a section, and it will change the days that the section meets on (even though I didn't add any days). This seems to happen when I have multiple sections for one course.

bryantgeorge (changes requested):
Okay, the above comments cover the extent of the things I noticed in testing. This is a really cool feature overall!

Notes:
- I spent most of my time testing and less time reading the code since there are so many lines changed.
- If any of the bugs I noticed are present here but not in #120 then let me know.

Jdellis818:
> This branch seems to cause blank values to be filled in with 0s. To replicate, make a course with a minimal number of fields filled in (prefix, number, section letter, room, days, duration, and instructor), then click on the section to edit, and it will have 0s in many of the previously blank fields, like so:
> 
> image

Yeah, I noticed this issue and figured out what caused it. Since nearly all of the fields in the Course, Section, and Meeting interfaces are required, the empty objects that are originally assigned to each section when the CSV is read, contain values of 0 for all the `number` fields. We fixed this before this PR by clearing any 0 values and interpreting them as empty, but that's not a good long-term solution since we should allow 0 as a possible value. In the future, we'll need to make most of the fields in the schedule interfaces optional, but that should be addressed in a different PR since it has a lot of ramifications especially in writing to the csv. I'll make an issue for it.

client-course-schedulizer/src/utilities/services/validation/popover.ts
Jdellis818: Positive doesn't work because it doesn't allow 0 as a possible value. I added integer because I thought it was better than nothing, but now realize that that doesn't work either since there can be fractional loads. I'll remove it, but I think we should figure out how to add a custom function to allow only positive numbers or 0.

Jdellis818:
> If multiple semesters are selected, should that multiply the value of faculty hours? (This might be another judgement call)

I don't think so based on how the template non-teaching loads spreadsheet was structured

client-course-schedulizer/src/utilities/services/validation/popover.ts
rpruim: Can you write a .nonnegative()?

Jdellis818: After some research, it looks like .min(0) should do the trick

charkour: :dab:

Jdellis818:
> It seems that you can only click the non-teaching loads to edit if the teacher also has a regular course (otherwise clicking the non-teaching load does nothing)

@bryantgeorge I can't seem to reproduce this. How did you get this bug to occur?

Jdellis818:
> Might be nice to switch the mouse cursor to the pointing hand icon when hovering over a clickable non-teaching load or section.

I added a pointer for the cursor when hovering over the table body. It's better than it was before, but stays as a pointer even for the cells that are not clickable. I think it will be somewhat difficult to detect whether the table cell is actually clickable or not. Do you think it's worth it to try and distinguish and switch the cursor between the pointer and normal?

bryantgeorge:
> > It seems that you can only click the non-teaching loads to edit if the teacher also has a regular course (otherwise clicking the non-teaching load does nothing)
> 
> @bryantgeorge I can't seem to reproduce this. How did you get this bug to occur?

I went to the teaching loads section, and added a teaching load for a professor which doesn't exist yet, so it looked something like:

image

and I wasn't able to click `Research (2)` unless I added a section for Jane Doe to teach.

bryantgeorge:
> > Might be nice to switch the mouse cursor to the pointing hand icon when hovering over a clickable non-teaching load or section.
> 
> I added a pointer for the cursor when hovering over the table body. It's better than it was before, but stays as a pointer even for the cells that are not clickable. I think it will be somewhat difficult to detect whether the table cell is actually clickable or not. Do you think it's worth it to try and distinguish and switch the cursor between the pointer and normal?

Something like this should work:
```diff
diff --git a/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx b/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
index 4abac7e..599329c 100644
--- a/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
+++ b/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
@@ -213,8 +213,14 @@ export const FacultyLoads = () => {
                     // Loop over the rows cells
                     row.cells.map((cell) => {
                       // Apply the cell props
+                      const cellClass = cell
+                        .getCellProps()
+                        .key.toString()
+                        .match(/(CourseSections)|(otherDuties)/g)
+                        ? "change-cursor"
+                        : "";
                       return (
-                        <TableCell {...cell.getCellProps()}>
+                        <TableCell className={cellClass} {...cell.getCellProps()}>
                           {
                             // Render the cell contents
                             cell.render("Cell")
```

Then only change the cursor for the `change-cursor` class

Jdellis818:
> > > Might be nice to switch the mouse cursor to the pointing hand icon when hovering over a clickable non-teaching load or section.
> > 
> > 
> > I added a pointer for the cursor when hovering over the table body. It's better than it was before, but stays as a pointer even for the cells that are not clickable. I think it will be somewhat difficult to detect whether the table cell is actually clickable or not. Do you think it's worth it to try and distinguish and switch the cursor between the pointer and normal?
> 
> Something like this should work:
> 
> ```diff
> diff --git a/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx b/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
> index 4abac7e..599329c 100644
> --- a/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
> +++ b/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
> @@ -213,8 +213,14 @@ export const FacultyLoads = () => {
>                      // Loop over the rows cells
>                      row.cells.map((cell) => {
>                        // Apply the cell props
> +                      const cellClass = cell
> +                        .getCellProps()
> +                        .key.toString()
> +                        .match(/(CourseSections)|(otherDuties)/g)
> +                        ? "change-cursor"
> +                        : "";
>                        return (
> -                        <TableCell {...cell.getCellProps()}>
> +                        <TableCell className={cellClass} {...cell.getCellProps()}>
>                            {
>                              // Render the cell contents
>                              cell.render("Cell")
> ```
> 
> Then only change the cursor for the `change-cursor` class

Awesome, thanks!

Jdellis818:
@bryantgeorge Thanks so much for the great review! I'm sorry for marking this PR ready for review so early. There were a lot of bugs, but you did a great job testing and catching a bunch that I missed! I think I've addressed all the issues which you brought up. I still couldn't reproduce the bug for professors with no classes though, so let me know if you're still experiencing that issue.

client-course-schedulizer/src/utilities/services/facultyLoadsService.ts
```diff
@@ -0,0 +1,279 @@
...
+  }
+  const [course] = courses;
+
+  const sections = filter(course.sections, (s) => {
+    return s.letter === letter && s.term === term;
+  });
+  if (!sections.length) {
+    return null;
+  }
+  const [section] = sections;
```
charkour, line 159: firstSection

client-course-schedulizer/src/utilities/services/facultyLoadsService.ts
```diff
@@ -220,3 +214,66 @@ export const getCourseSectionMeetingFromCell = (
...
+    return (
+      s.isNonTeaching === true &&
+      s.instructionalMethod === nonTeachingLoad &&
+      s.instructors.includes(instructor)
+    );
+  });
+  if (!sections.length) {
+    return null;
+  }
+  const [section] = sections;
```
charkour: Possibly use more explicit variables names here that the course is the non-teaching loads and section is an individual non-teaching load.

client-course-schedulizer/src/components/Tabs/FacultyLoads/UpdateNonTeachingLoadModalPagination/UpdateNonTeachingLoadModalPagination.tsx
```diff
@@ -0,0 +1,93 @@
...
+        onClose={handleModalClose}
+        onPageChange={handlePageChange}
+        open={open}
+        page={page}
+        pageCount={instructorNonTeachingLoads.nonTeachingLoads.length}
+      >
+        <AddNonTeachingLoadPopover values={modalValues} />
+      </ModalPagination>
+    );
+  },
```
charkour, line 21: This is similar to the UpdateSectionModalPagination. I'm trying to think if it would be worth it to make one component more generic, or if it's fine to leave these separate but similar.

charkour (changes requested):

client-course-schedulizer/src/components/Tabs/FacultyLoads/UpdateNonTeachingLoadModalPagination/UpdateNonTeachingLoadModalPagination.tsx
charkour: I'm trying to see if we can combine them.

bryantgeorge:
Not a big deal, but this will make it so that blank cells won't look clickable:

Applied in 06c2eca

```diff
diff --git a/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx b/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
index d9cb67f..f965d2d 100644
--- a/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
+++ b/client-course-schedulizer/src/components/Tabs/FacultyLoads/FacultyLoads.tsx
@@ -133,12 +133,14 @@ export const FacultyLoads = () => {
                       // Loop over the rows cells
                       row.cells.map((cell) => {
                         // Apply the cell props
-                        const cellClass = cell
-                          .getCellProps()
-                          .key.toString()
-                          .match(/(CourseSections)|(otherDuties)/g)
-                          ? "change-cursor"
-                          : "";
+                        const cellClass =
+                          cell.value &&
+                          cell
+                            .getCellProps()
+                            .key.toString()
+                            .match(/(CourseSections)|(otherDuties)/g)
+                            ? "change-cursor"
+                            : "";
                         return (
                           <TableCell
                             className={cellClass}
```

bryantgeorge:
> I still couldn't reproduce the bug for professors with no classes though, so let me know if you're still experiencing that issue.

I wasn't able to reproduce this when I tried today, so seems that it's been fixed (or I was just doing something wrong in the first place)

charkour (approved):
This is good to be merged after some better documentation is added.

PR #137, opened 2021-03-05
Add FieldArrayForm Component

### LINKED ISSUE #136: Work on Harmony integration
This is a generic issue to link all harmony related PRs to. Not sure what each piece of Harmony will be, kinda seeing what seems natural.

charkour:
Don't feel like you need to spend a lot of time with this. I made this PR so you guys can be aware of how Harmony is laid out and what components you have access to. 

In this PR: 

- Move Harmony to a page rather than a tab
  -   _Why?_ Harmony is independent of Schedulizer's data but the two can send data back-and-forth to each other (not fully implemented yet). User uploads a schedule to Schedulizer --> uses Harmony to make a schedule from pre-filled data --> finds a working schedule and sends it to the Schedulizer to continue making tweaks. 
- Create the FieldArrayForm component and sub-components

A little more details on the FieldArrayForm. This component allows for forms that are arrays of various lengths which is important when editing data relating to Harmony. To keep things simple, I've split the FieldArrayForm into multiple components and created a provider to pass data down the component tree to the necessary components. 

To Test:
- `npm i && npm start`
- Click on Harmony tab
- Fill in forms and click submit (output logged to console). 

Any questions or critiques would be great! Links to #136


Jdellis818 (approved):
LGTM! This is definitely a great start on the UI for Harmoniously. The one concern I have is with all the variables being in separate forms. You probably will want each of them in the same form with just 1 submit button at the bottom, right? Also, you probably want the add button to be singular rather than plural since you're only adding one at a time (i.e. "Add Professor" instead of "Add Professors")

charkour:
Yeah, that is right. I'll work those two changes into the next PR. Thanks!

PR #144, opened 2021-03-09
combine popover modal logic

charkour:
I talked to Jon about making the ModalPagination a little more generic so we didn't have to repeat similar logic. This moves logic out of the UpdateNonTeachingLoadModalPagination and UpdateSectionModalPagination components and places it in ModalPagination. Now those previous components only need to provide the ref, the search function for the CSM, and what popover component to show. 

Also, I settled on the verbiage "iterable" and "key" but I am open to suggestions.

Overall, it is more confusing, but lines aren't repeated. So I am not sure if it is worth it to merge this in, but now we have the options if we would like. We can add comments to the code if we decide this PR is worth it. 

131 additions and 248 deletions. 

I'm wondering if the package-lock.json additions are because we are on different node versions. I am on v14.15.1.

### DIFF
--- a/client-course-schedulizer/package-lock.json
+++ b/client-course-schedulizer/package-lock.json
@@ -1,8 +1,21411 @@
 {
   "name": "client-course-schedulizer",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "0.1.0",
+      "dependencies": {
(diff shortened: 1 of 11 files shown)


bryantgeorge (approved):
Generally, I think this is an improvement and makes things more modular, LGTM

charkour:
Thanks. I'll check on another branch to see if that's a bug before merging this.

charkour:
Yeah, this bug is not due to this PR

PR #149, opened 2021-03-10
persist tab state

charkour:
A little PR to keep the Schedulizer Tab state stored in localStorage to persist throughout sessions. Helps make the app easier to predict from the user's point of view. 

Thanks for the comments and critiques.


Jdellis818 (approved):
LGTM! Great work!

PR #150, opened 2021-03-14
Scroll to Section

### LINKED ISSUE #132: Sort Profs by First Name for now

bryantgeorge:
To Test:
- Try updating sections, adding section from an existing section, adding completely new sections, and adding non-teaching loads (where applicable try doing so on both the faculty schedule, room schedule, and faculty loads tabs to make sure it works everywhere)
- Upon performing any of these actions, it should scroll to the section/load which was added/updated (or the instructor's row if you are on the faculty loads tab)
- Notice that instructor names and rooms are now sorted alphabetically
- Probably should try changing a section from one term to another (this should switch terms and then scroll to the section)

Notes:
- If there are multiple rooms/instructors for a given section, the scrolling will need to make a choice about which to scroll to, and it may not end up choosing the one you clicked (tries to choose the last one)
- The faculty loads page will scroll that instructor's row to the center of the screen (or close as close to the center as possible), but the room/faculty schedule tabs will just scroll that section into view because if I try to scroll to the center, it causes graphical issues

Closes #150 
Also Closes #132


client-course-schedulizer/src/utilities/hooks/addSectionHooks.ts
```diff
@@ -61,16 +84,60 @@ export const useAddSectionToSchedule = () => {
...
+      const room = `${meeting.location.building}_${meeting.location.roomNumber}`;
+      className = createEventClassName(newSectionName, room, prof);
+      const newElements = document.getElementsByClassName(className);
+      if (newElements) {
+        newElement = newElements.item(newElements.length - 1) ?? undefined;
+      }
+    });
+  });
+  if (newElement && newElement.parentElement?.parentElement) {
+    newElement.scrollIntoView({ behavior: "smooth", inline: "nearest" });
```
Jdellis818: I tried changing this to `inline: "center"` and it seemed to work as expected. What graphical issues did you experience?

Jdellis818 (changes requested):
Nice work! This looks good to me. The implementation is very clean. I just have a few questions to think about before merging this in. Is there a way to keep the page from scrolling to the bottom before scrolling over? It's not a huge deal, but at least on my small screen, I have to scroll up again to see a section once it is moved. Also, I think it's fairly important to scroll the schedule to the center if possible. Again, it's not a huge deal but it could be a little frustrating to need to scroll more after the automatic scroll. This PR will also increase the priority of sorting the faculty loads table since I'm sure the user will want to sort by total hours a bit of the time.

client-course-schedulizer/src/utilities/hooks/addSectionHooks.ts
bryantgeorge: So to test the graphical issues, set it to `inline: "center"` like you did, import the math schedule (or another bigger schedule), and edit a section for the last teacher listed (rightmost). When it scrolls over to that section once updated, it ends up shifting all of the content of the site to the left leaving some blank space on the right and hiding some of the content on the left.

Are you able to reproduce this behavior? (I was testing on Chrome in case that's relevant)

bryantgeorge:
@Jonri2 thanks for the feedback!

I added sorting on columns in the faculty loads tab (when clicked) although it looks a little rough, so if you have suggestions on styling, let me know.

I also made it so that it shouldn't scroll the calendar event to the center vertically anymore.

Finally, see my comment about horizontal centering issues, and let me know if you can reproduce the issue.

client-course-schedulizer/src/utilities/hooks/addSectionHooks.ts
Jdellis818: Yes, I see the issue now. Let's just keep it as is then.

Jdellis818 (approved):
LGTM. We'll have to make the columns of the loads table look sortable to the user since the arrows don't appear until after you click, but that can be addressed later since it's mostly out of the scope of this PR. Nice work!

PR #151, opened 2021-03-15
Fix fractional faculty loads bug and add total hours to table

Jdellis818:
This PR fixes the bug where fractional faculty loads are set to 0. It also fixes a bug where faculty loads for sections with multiple professors were miscalculated and adds a total to the loads table. To test, import a schedule with fractional faculty loads. It should successfully compute the loads in the table and be successfully added to an exported spreadsheet.

PR #153, opened 2021-03-17
Fix non-teaching load export bug

### LINKED ISSUE #142: Fix non-teaching activity load hour handling.
There's a bug that messes up the non-teaching load hours. If you:
1. Create a new non-teaching activity (e.g., Dept-Chair, 7 hours). This displays properly in the teaching-loads pane.
2. Export to draft. This saves the entry properly at the end of the CSV.
3. Import the saved draft. Now, the chair entry is gone.

The one-hour non-teaching activities seem to be working fine.

Jdellis818:
This fixes the issues specified in #142. To test, import a schedule and create many non-teaching load activities (more than 1 for a professor, some with no term, etc.). Then, export and reimport the schedule. The non-teaching loads should work exactly as they did before the export. I believe the issue with all the non-teaching loads importing with the same load hours was fixed with #151.

### DIFF
--- a/client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
+++ b/client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
@@ -80,7 +80,7 @@ export const termCallback = (value: string, { section }: CaseCallbackParams) => 
     });
     section.term = termsArr;
   } else {
-    section.term = termCase(value);
+    section.term = value === "" ? [] : termCase(value);
   }
 };
 
@@ -99,7 +99,7 @@ export const instructorCallback = (value: string, { section }: CaseCallbackParam
 };
 
 export const prefixCallback = (value: string, { course }: CaseCallbackParams) => {
-  course.prefixes = prefixCase(value);
+  course.prefixes = value === "" ? [] : prefixCase(value);
 };
 
 export const nameCallback = (value: string, { course }: CaseCallbackParams) => {
--- a/client-course-schedulizer/src/utilities/helpers/writeCSV.ts
+++ b/client-course-schedulizer/src/utilities/helpers/writeCSV.ts
@@ -74,6 +74,9 @@ export const scheduleToNonTeachingCSVString = (schedule: Schedule): string => {
 
 export const getTermsStr = (section: Section): string => {
   if (Array.isArray(section.term)) {
+    if (section.term.length === 0) {
+      return "";
+    }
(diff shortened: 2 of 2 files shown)


charkour (approved):
LGTM. Was able to test it and it seems to work.

@kvlinden this will be in develop shortly.

kvlinden:
@Jonri2 
I just tried this on the development branch and I'm now able to create/display/sum multiple non-teaching activities of different hour loads per faculty, but only the first activity is exported per faculty member, and if I manually add a CSV row back in for a second activity for that faculty member, only the first is imported.

Am I using the merged changes?

Jdellis818:
@kvlinden I just tested it on develop and am not able to reproduce that issue. It could be that there are issues with the spreadsheet you are importing if it's an export from before the bug was fixed. Could you try again with a schedule that starts with no non-teaching loads?

charkour:
@kvlinden, I believe in order to get the updates, you need to refresh your website once in order for the newest javascript to load. We could consider adding the commit hash to our website using GitHub actions so we can see which version is on develop so you can be sure it's on the latest. Or we could alert the user when they are on an older version.

kvlinden:
@Jonri2 & @charkour : I just tried again and got the same problem. 

On the development site (https://sharp-babbage-a45ee2.netlify.app/#/), I loaded a raw dump of the CS schedule, added two non-teaching activities for myself, which displayed/summed/exported properly. Here's the export file (converted to XLSX to make GitHub happy):

[full_schedule_2021-03-18T07 19 15-04 00.xlsx](https://github.com/senior-knights/course-schedulizer/files/6163603/full_schedule_2021-03-18T07.19.15-04.00.xlsx)

When loading (the CSV version), I only see the first activity.

Jdellis818:
@kvlinden Alright, I think I figured out what the bug is. It looks like only the 1st non-teaching load is taken for a professor if they are listed under the same term. I should have it fixed by the end of the day.

kvlinden:
Thanks - maybe it's the semester setting. A while back, I found that I needed to set a semester to get this to save, so I've just always done that. Actually, these non-teaching assignments don't have semesters, so it would make sense to leave the FA/SP/... un-specified, if that would help.

rpruim:
I think the deans to like to have these things declared in a particular semester.  And some things like sabbaticals and CRFs do have a specific semester in which the work is to be done.

PR #154, opened 2021-03-17
Set name and department on update

### LINKED ISSUE #145: Updating the Name field does not work
**Describe the bug**
Updating the Name field for a section/course does not work (change doesn't persist).

**To Reproduce**
1. Load a schedule
2. Click on a section
3. Change its `Name` field to something else
4. Click `Update Section`
5. Click that section again, and notice that the `Name` field is back to what it was originally (before you updated it)

Jdellis818:
Fixes a bug where name and department would not update. To test, click on a section and edit its name and department. Then, click on the same section again or any section from the same course and the changes should have persisted.

### DIFF
--- a/client-course-schedulizer/csv/math-schedule-export.csv
+++ b/client-course-schedulizer/csv/math-schedule-export.csv
@@ -50,8 +50,8 @@ T","10:30AM - 11:20AM
 10:30AM - 11:20AM","Adv Linear Algebra","Todd Kapitula"
 19/FA,"MATH-359-A",3.00,2.00,"NH 294","TH","6:00PM - 9:00PM","Sem: Sec Teaching of Math","David Klanderman"
 19/FA,"MATH-361-A",4.00,4.00,"NH 251","MTWF","12:30PM - 1:20PM","Real Analysis I","John Ferdinands"
-20/IN,"MATH-390-A",3.00,0.00,"","","","Modeling of Angiogenesis","Todd Kapitula"
-19/FA,"MATH-390-A",1.00,0.00,"","","","Modeling of Angiogenesis","Christopher Moseley"
+20/IN,"MATH-390-A",3.00,0.00,"","","","Galois Theory","Todd Kapitula"
+19/FA,"MATH-390-A",1.00,0.00,"","","","Galois Theory","Christopher Moseley"
 20/SP,"MATH-391-A",0.00,1.00,"NH 276","TH","3:30PM - 4:20PM","Colloquium","David Klanderman
 Christopher Moseley"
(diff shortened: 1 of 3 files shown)


charkour:
Thanks for this. Looks like there is an issue with the tests. Probably due to the changes to the test files in a previous PR.

Jdellis818:
No. It's because our test file has 2 sections of MATH-390 with different names and this PR makes it take the 2nd one instead of the 1st one. Both ways are incorrect though. It should have a different name for each section. I'm not really sure how to deal with that. Any ideas?

charkour:
I don't have any initial thoughts. That's an interesting problem. I can look into it later.

Jdellis818:
> I don't have any initial thoughts. That's an interesting problem. I can look into it later.

Yeah, I think that it is incorrect to have 2 different names for the same course in which case it's just the spreadsheet that is wrong, but we can ask Prof. Pruim what behavior he would expect.

rpruim:
FYI: 390 is an interesting thing: Eventually it gets two instructors.  But since that looks like a conflict, I assign one each semester as I'm creating the schedule and then I let the registrar know it is both both semesters.  That's a work-around, not ideal behavior.

When you say name do you mean course title?

rpruim:
A course like Math 385 (Topics in Mathematics) could have multiple titles.  Next fall it will be called Topics in Mathematics: Optimization.  But it will have different names each time it is offered.  But generally, the title should be the same each time.  Mismatches likely are caused from sloppy typing or from forgetting to update a title when modifying a course number as part of schedule construction.

Jdellis818:
> FYI: 390 is an interesting thing: Eventually it gets two instructors. But since that looks like a conflict, I assign one each semester as I'm creating the schedule and then I let the registrar know it is both both semesters. That's a work-around, not ideal behavior.
> 
> When you say name do you mean course title?

Yes. One is called Modeling of Angiogenesis and one is called Galois Theory, but our interface currently only allows 1 title for a course so it just picks one of them on import and then both are exported with the same name.

charkour:
Would it be better if we allow multiple course names to be separated by a semicolon as we do for professors?

charkour (changes requested):
Would it be better if we allow multiple course names to be separated by a semicolon as we do for professors? Would it be better to add it to a different PR?

Jdellis818:
> Would it be better if we allow multiple course names to be separated by a semicolon as we do for professors? Would it be better to add it to a different PR?

Perhaps, though that may be confusing. Let's just stick with this and make a low priority issue

charkour:
Or I'm thinking it maybe be better to assign a name on a per semester basis? Regardless, I'll make an issue and we can figure it out later.

charkour (approved):
LGTM. 

That's an important two lines of code.

PR #156, opened 2021-03-18
Add instructional method to existing section check

Jdellis818:
This should fix the bug where some non-teaching loads are dropped on import if a professor has multiple non-teaching loads for the same term. Instructional method is important because it is the field where the activity name of a non-teaching load is stored. The issue before was that 2 non-teching loads for the same term and professor were looking like a duplicate and being removed. By checking instructional method as well, 2 non-teaching loads will not be considered duplicates unless they have the same instructor, term, and activity. Please test this a bunch to make sure there are no more issues.


charkour (approved):
Looks good to me. I was able to create teaching loads there were dropped on our current develop version that were properly imported on this branch.

PR #159, opened 2021-03-21
Edit multiple meetings support

### LINKED ISSUE #139: Add support for sections with multiple-room/time schedules
The system doesn't handle a course section with multiple rooms reserved for different time periods. DATA 101 B, Fall, 2021 is an example. It reserves NH 253 (MWF, 3:30-4:30) and SB 354 (MW, 3:30-5:20). Here's the CSV entry from Colleague:

Computer Science | 21/FA | ######## | 2021 | DATA-101-B | DATA | 101 | B | 100 | 3 | 3 | 0 |   | 32 | 32 | 34 | NH 253     SB 354 | MWF     MW | 3:30PM - 4:20PM     3:30PM - 5:20PM
  -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

Editing any one of the section instances of this course in Schedulizer blows away the other classroom booking. The only workaround for this that I could find is to edit the CSV (to reserve both rooms at the same times for that section) before uploading it to Schedulizer, If we really wanted the two-hour bookings on MW, then we'd have to add another DATA 101 lab section of some sort.

This is a pathalogical case. If it's hard to fix, we could simply fix it manually the first time and promise not to submit such a course in future.

### LINKED ISSUE #157: Remove final export
Just allow the draft (full) csv

bryantgeorge:
To Test:
- Notice that there is only one export button now (for the full/draft version of the csv)
- Create a section
- Click that section, and change some subset of the start time, duration, days, and/or room, then click `Add New Section`
- Now export and notice that there is only one entry in the CSV (which contains both meetings (or meeting sets might be a more accurate term))
- Try editing one of the meetings, export again, and ensure that your changes only applied to the meeting that you clicked on
- Click `Delete Section` for one of the meetings, and ensure that it only deleted that meeting (then you can try deleting the other, and that should delete the course and section itself)
- Try out other edge cases, like adding a meeting for an existing section in the math schedule, updating a course/section like normal, non-teaching loads still work, etc.

Notes:
- This is a pretty major change to how our app works, so it should be treated with some caution in testing
- Is it confusing that pressing `Delete Section` won't delete the whole section if there are multiple meetings in that section? Should we call this button, `Delete Meeting` or something else?
- This will unfortunately make it so that changes which apply on the section level must be made for each meeting in the section individually (so if you had 2 meetings for a section taught by Bob and wanted to change the section to be taught by Sarah, then you would need to do it for both meetings). We may want to brainstorm ways to make section/course-wide edits in the next meeting.

Closes #139
Also Closes #157


Jdellis818:
If this PR is also handling multiple rooms for a section then #162 may slightly change the way this is handled since multiple autocomplete will return an array rather than a string separated by commas.

bryantgeorge:
> If this PR is also handling multiple rooms for a section then #162 may slightly change the way this is handled since multiple autocomplete will return an array rather than a string separated by commas.

Thanks for the heads up, I don't think this will cause much conflict because this is primarily about multiple meetings, and your changes are about multiple instructors/rooms being assigned to a given meeting correct?

Jdellis818 (changes requested):
It's pretty cool that you were able to get this working for the most part. It's definitely a difficult problem and it's awesome that you were able to figure it out. After testing a bit, I did uncover a fairly major issue. A section is no longer able to update or be added if the section letter is changed. So, if I open an existing section and change the letter to something new, then click on either "Update Section" or "Add New Section," nothing happens. Maybe there should be a separate "Add New Meeting" button? I'm unsure what the best solution is here. Also, I think it may be best to still export non-teaching loads as a separate spreadsheet.

bryantgeorge:
@Jonri thanks for noticing that bug. It turns out that it's also on `develop`, and I found some other fairly significant bugs that I made an issue for: #167

I'll start working on that issue now.

Jdellis818 (approved):
LGTM. The only thing I would say is maybe only export the non-teaching loads spreadsheet if there are non-teaching loads in the schedule.

PR #162, opened 2021-03-23
Add Autocomplete to Form Fields

### LINKED ISSUE #112: add autofill for names, room numbers, etc.

Jdellis818:
This PR implements autocomplete for the Add Section Form and the Add Non-Teaching Load form. To test, click on a section. Number, Section, Name, Instructional Method, and Room have regular autocomplete and Prefix and Instructor have multiple autocomplete (meaning more than 1 value can be selected). You should be able to enter a value that is not an option in any of the autocomplete fields and it should work as expected. At the moment, multiple rooms are not supported but I anticipate fixing that in a future PR. Let me know what you guys think!


client-course-schedulizer/src/utilities/services/addSectionService.ts
```diff
@@ -40,13 +34,13 @@ export interface SectionInput {
...
   instructionalMethod: Section["instructionalMethod"];
-  instructor: Instructor;
+  instructor: Instructor[];
   intensiveSemester: Intensive;
   localMax?: Section["localMax"];
   location: string;
   name: Course["name"];
   number: Course["number"];
-  prefix: Prefix;
+  prefix: Prefix[];
```
charkour, line 43: What are the implications of making this change to arrays? Will this affect how we parse things when importing and exporting schedules?

charkour (changes requested):
I have one question and also when I add multiple professors, I can't click submit because the popover goes off of the screen. 

Nice work on the autocomplete component! It looks good and works well.

client-course-schedulizer/src/utilities/services/addSectionService.ts
Jdellis818: There shouldn't be any implications since prefixes is already an array in the `Course` interface. This change is just for the form input interface and makes it so you know longer have to split on commas.

Jdellis818:
> I have one question and also when I add multiple professors, I can't click submit because the popover goes off of the screen.

Hmm... on my machine there's a scroll bar so I can just scroll down when there are multiple professors. Do you have any suggestions on how to stop it from going off the screen?

charkour:
Initially, I'm not sure. Is there a scroll bar on the popover form?

Jdellis818:
> I'm not sure. Is there a scroll bar on the popover form?

Yes

charkour:
Okay, I'll check it out and see if I can get the scroll bar to show up.

Jdellis818:
```
diff --git a/client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx b/client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx
index af9e3c8..dc29801 100644
--- a/client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx
+++ b/client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx
@@ -119,7 +119,7 @@ const ScheduleBase = ({ calendarHeaders, groupedEvents, ...calendarOptions }: Sc
           horizontal: "left",
           vertical: "bottom",
         }}
-        PaperProps={{ style: { maxWidth: "90%", minWidth: "500px" } }}
+        PaperProps={{ style: { maxWidth: "90%", minWidth: "500px", overflowY: "scroll" } }}
         transformOrigin={{
           horizontal: "right",
           vertical: "top",
```
Does this work?

charkour:
No, it does not fix it :(

charkour:
It works if you add a maxHeight property to the PaperProps. overflowY: scroll does not need to be specified.

Jdellis818:
Alright, you should hopefully be good now

client-course-schedulizer/src/components/reuseables/Schedule/Schedule.tsx
```diff
@@ -119,7 +119,7 @@ const ScheduleBase = ({ calendarHeaders, groupedEvents, ...calendarOptions }: Sc
           horizontal: "left",
           vertical: "bottom",
         }}
-        PaperProps={{ style: { maxWidth: "50%", minWidth: "500px" } }}
+        PaperProps={{ style: { maxHeight: "95%", maxWidth: "90%", minWidth: "500px" } }}
```
charkour: With 95%, it still gets cut off. Sorry that there is such weird behavior with MacOS. 90% should work. Thanks for the fixes.  

```suggestion
        PaperProps={{ style: { maxHeight: "90%", maxWidth: "90%", minWidth: "500px" } }}
```

charkour (changes requested):
You can merge it once the tests pass.

PR #165, opened 2021-03-24
Add timestamp

### LINKED ISSUE #119: Time stamp last change for each section?
I'm wondering if it would be useful to add an additional column that keeps track of the last change to a section.  Ideally this would only update when there is an actual change, not each time the section is "updated".

This can be useful for keeping track of changes -- it would be easy to filter by date to get recent changes.

For now I'm imagining that we would only store the time stamp -- we can use external tools to do the filtering.

Jdellis818:
This PR adds a field called LastEditTimestamp to the export so that users can sort the spreadsheet by the timestamp to see which sections were most recently edited.

To Test:
- Upload a schedule
- Update a few of the sections
- Export the full spreadsheet
- Open the spreadsheet in Excel and sort by the LastEditTimestamp field. Your edited sections should be at the top
- Reimport the spreadsheet and edit a few different sections
- Export and sort again. Notice that your most recent changes are on top, then the changes from the previous export, then the unchanged sections

Closes #119


bryantgeorge:
Do we want to do this in the non-teaching loads export as well? (make sure to merge `develop` in before trying this)

bryantgeorge:
This isn't a big deal, and is more of a question than a suggestion: do we want to update this field if nothing else about the course changed? (ie. if you make no edits but press the update button, should the time updated change?)

bryantgeorge (changes requested):
Looks great other than the two minor issues above.

charkour:
> This isn't a big deal, and is more of a question than a suggestion: do we want to update this field if nothing else about the course changed? (ie. if you make no edits but press the update button, should the time updated change?)

If it's not hard to add that logic, then I think the time shouldn't be updated if no fields are changed. But if it's hard, then don't add that feature since if a user does not want to make a change, they can click out of the popover.

bryantgeorge:
> Do we want to do this in the non-teaching loads export as well? (make sure to merge `develop` in before trying this)

Do we want these in just the main csv which is exported, or do we want to add these to the non-teaching csv as well?
Otherwise this looks great!

Jdellis818:
@bryantgeorge I just addressed your changes :). I made it so it actually won't go through any of the submit logic if none of the fields are updated. That was a lot harder than I thought it'd be haha. I also added the timestamp to non-teaching loads in the main csv, but not in the non-teaching loads csv. I can add it in the non-teaching loads csv if you think that's better.

bryantgeorge (approved):
LGTM, I think I would suggest putting the timestamps in the non-teaching csv as well (but not a big deal if you disagree)

PR #166, opened 2021-03-24
Make schedule interface fields optional

### LINKED ISSUE #135: Make Fields in Schedule Interfaces Optional
Fields that are not required when adding a section should be optional.

### LINKED ISSUE #155: Handle multiple names for a course
Not sure what the best way is that. Reference the conversation on #154

Jdellis818:
This PR makes most of the schedule fields optional. This allows optional fields such as anticipated size, global max, used, etc. to be blank rather than defaulted to 0 in the add section form and the export of the schedule. It also fixes #155 and an issue where non-teaching loads would be duplicated when updating.


bryantgeorge (changes requested):
I found an issue that I don't think is directly related to these changes, but if you fill in a field like `Name` and then clear it and try to add the section, it will give an error like this:

image

Also this unfortunately pushes the `Add Section` button down such that you can't see it anymore (though if we fix this big, maybe that won't be an issue)

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
```diff
@@ -127,17 +127,17 @@ export const anticipatedSizeCallback = (value: string, { section }: CaseCallback
...
 
 export const yearCallback = (value: string, { section }: CaseCallbackParams) => {
   section.year = yearCase(value);
 };
 
 export const studentHoursCallback = (value: string, { section }: CaseCallbackParams) => {
-  section.studentHours = value.startsWith("$")
-    ? integerDefaultZeroCase(value.substr(1))
-    : integerDefaultZeroCase(value);
+  section.studentHours = value.startsWith("$") ? Number(value.substr(1)) : Number(value);
```
charkour, line 141: why do we check for $? perhaps adding a comment would be good since it isn't immediately apparent.

client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
```diff
@@ -349,6 +350,6 @@ export const durationCase = (value: string): number => {
   return endTimeMoment.diff(startTimeMoment, "minutes");
 };
 
-export const yearCase = (value: string): number | string => {
-  return Number.isInteger(Number(value)) ? Number(value) : value;
+export const yearCase = (value: string): number | string | undefined => {
+  return Number.isInteger(Number(value)) && value ? Number(value) : value;
```
charkour: Does this work? if `value` is falsy, then it will short circuit and not try to cast to a number. Added parens to make it more explicit what the precondition is for the ternary operator. 

```suggestion
  return (value && Number.isInteger(Number(value))) ? Number(value) : value;
```

client-course-schedulizer/src/utilities/helpers/readCSV.ts
```diff
@@ -155,8 +155,10 @@ export const insertSectionCourse = (schedule: Schedule, section: Section, course
       section.instructionalMethod,
     );
 
-    schedule.courses[existingCourseIndex].name = course.name;
     schedule.courses[existingCourseIndex].department = course.department;
+    if (course.name !== schedule.courses[existingCourseIndex].name) {
```
charkour, line 162: is this checking if a course name is already present in the list of course names?

client-course-schedulizer/src/utilities/interfaces/dataInterfaces/scheduleInterfaces.ts
```diff
@@ -21,47 +21,46 @@ export interface Meeting {
...
 export interface Section {
   anticipatedSize?: number;
-  comments: string;
+  comments?: string;
   // Number of students enrolled in this section 10 days into the course
   day10Used?: number;
   // Like 2/3/2020
-  endDate: string;
+  endDate?: string;
   // Overrides Course value
```
charkour: remove this comment since it doesn't override the course value anymore (facultyHours isn't on the course datatype anymore).

client-course-schedulizer/src/utilities/interfaces/dataInterfaces/scheduleInterfaces.ts
```diff
@@ -21,47 +21,46 @@ export interface Meeting {
...
   meetings: Meeting[];
-  semesterLength: SemesterLength;
+  name?: string;
+  semesterLength?: SemesterLength;
   // Like 5/21/2020
-  startDate: string;
-  status: string;
+  startDate?: string;
+  status?: string;
   // Overrides Course value
```
charkour: remove this comment also

charkour (changes requested):
Outside of Bryant's comments, I just had some code clarity questions to make sure I am following what is happening. The majority of the code looks really solid! It makes the app more robust. Nice work.

I have not tested this extensively.

charkour (approved):
LGTM. Just fix the merge conflicts and then it is good to go.

Issue #167, opened 2021-03-25
Section entry bugs


bryantgeorge:
This is very high priority!

1. From #159, @Jonri2 noticed that: "After testing a bit, I did uncover a fairly major issue. A section is no longer able to update or be added if the section letter is changed. So, if I open an existing section and change the letter to something new, then click on either "Update Section" or "Add New Section," nothing happens." Turns out this is in `develop` as well. After a bit of debugging, the issue is that the updated section letter isn't even sent to `addSectionToSchedule()`.
2. Making a non-teaching load with multiple teachers makes a new row in the table, like:
![image](https://user-images.githubusercontent.com/51130302/112473088-4cc1e480-8d44-11eb-8f98-fd2d35144499.png)
3. If you enter a new prefix, and you don't press enter, it creates undefined values (validation no longer ensures there is a prefix):
![image](https://user-images.githubusercontent.com/51130302/112473645-fd2fe880-8d44-11eb-9229-d65f4f3d1d21.png)

(Related to number 3, removing all instructors from a course, removes it from the room schedule even though it should still appear there, correct?)

Jdellis818:
2 and 3 were definitely caused by #162. The validation should be an easy fix. I accidentally put `.required()` inside of the `array()` instead of outside

bryantgeorge:
> 2 and 3 were definitely caused by #162. The validation should be an easy fix. I accidentally put `.required()` inside of the `array()` instead of outside

Yeah, that one seems easy (the error message also needed to be added to the griditemautocomplete)

PR #168, opened 2021-03-25
Fix some issues with autocomplete

Jdellis818:
This PR fixes 1 and 2 of issue #167. The first was fixed simply by adding the autoSelect prop to the Autocomplete which adds the current value when the Autocomplete loses focus. The second was fixed by making `instructor` an array in `NonTeachingLoadInput`. It also fixes a bug where the instructor was not passed to the non-teaching load form when editting.


bryantgeorge (approved):
Thanks a ton for finding a fix for both of these! AutoSelecting the entered text is really nice.

I added some commits which:
1. Merge in `develop` (there weren't any conflicts but I wanted to be sure)
2. Fixes number 3 from #167 (hopefully)
3. Fixes a bug I found where leaving the instructor/prefix field without entering anything while there is already an instructor/prefix in the field would delete an existing instructor/prefix (because the already entered thing is highlighted in the selector, and if you enter something already there, it removes it for some reason). The way I fixed this was by removing the instructors/prefixes already in the text field from the autocomplete options (and I removed the empty string which is confusing to include).
4. I made it so that the autocomplete options appear whenever the field gains focus (previously if you tabbed into the field, the autocomplete options would not appear)

Feel free to look over and test my changes, and if you don't have any concerns, you can merge

PR #169, opened 2021-03-26
Add Red Border Around Conflicts

### LINKED ISSUE #127: Make conflicts easier to find.
Adding a small border would potentially do it without writing logic to check conflicts might be good enough.

Jdellis818:
This PR adds a red border for all meetings that have conflicting instructors or rooms. 
To test:
- Import the math-schedule. It should have a red border around the MATH-270 and MATH-271 sections since they look like a conflict.
- Change the color by and make sure the border stays
- Add a new section that conflicts with a professor's schedule. It should have a red border once added. 
- Check the room schedule and notice that the section you added also has a red border there (so users can see if there are room conflicts from the faculty view or instructor conflicts from the room view)
- Update it to a time that does not conflict and notice that the border goes away
- Test some of the edge cases such as multiple instructors, same time but different days, same time but different term, etc.

The next step will be to add this information to the Conflicts tab.

### DIFF
--- a/client-course-schedulizer/package-lock.json
+++ b/client-course-schedulizer/package-lock.json
@@ -28,6 +28,7 @@
         "material-ui-image": "3.3.1",
         "material-ui-popup-state": "^1.7.0",
         "moment": "2.29.1",
+        "moment-range": "^4.0.2",
         "node-sass": "4.14.1",
         "object-hash": "^2.1.1",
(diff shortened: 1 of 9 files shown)


charkour (approved):
Looks good to me! 

A clean implementation. Good comments and clear to understand.

Issue #143, opened 2021-03-08
Add edit button to faculty loads table


Jdellis818:
Adding a button will make it clearer that clicking on a table cell will not pull up a specific course, but the 1st course with pages for each one. 

Mentioned in #120

PR #171, opened 2021-03-29
Feature/teachingload item format

### LINKED ISSUE #140: Upgrade format of non-teaching activity entries.
The text presentation of non-teaching activity entries on the teaching-loads panel should:
- include the semester, e.g., karnold has a two-part CRF: 3 hours in spring; 3 in interim. This hours split is important for accounting and should, thus, be displayed.
- list each non-teaching activity on a separate line, which would make them easier to read.

kvlinden:
This (simple, practice) PR reformats the section & non-teaching duty entries on the Teaching Loads page to include the hours for each item. 
To test, load the Teaching Loads tab and check that the correct hours are listed for all sections and/or non-teaching duties.


bryantgeorge (changes requested):
These look better, but this breaks the click-ability of the sections/non-teaching loads. Probably will need to apply a regex to some of the functions later on in `facultyLoadsService.ts` to ignore the ` (#)` from the end of the strings.

kvlinden:
Sorry to miss that error. I've now parsed out the added load hours spec so that the popup works as it did before. 

I'll have to say that concatenating and then parsing these strings is a bit clunky. Note, also, that the popup only works on the first course/duty on the list (per #143). I'd guess that supporting the popup for any course/duty on the list (#120) will require creating a span for each course/duty on the list rather than treating them as a single string.

Jdellis818 (approved):
LGTM. It's fun to see you start to dive into the code :)

Jdellis818:
> Note, also, that the popup only works on the first course/duty on the list (per #143).

You can access any of the sections through the paginator at the top of the popup. It's a bit weird that the first is always brought up at first no matter which you click on, but there is a way to access any of them. #143 suggests an edit button for each course which would fix the weirdness of the UI.

Jdellis818 (changes requested):
After reviewing a little further it looks like that regex does not parse out fractional loads. So, the popup will not come up for a course with a fractional load.

kvlinden:
So, perhaps this wasn't such a "(simple, practice)" PR after all. Either this third time's a charm or charkour can take his turn at finding a bug. :-)

The regex now handles float or int load values, and only matches before commas or end-of-string. The user can still mess things up by putting a fake load spec with comma in a duty name. There may be a way to prevent this as well.

Jdellis818 (approved):
LGTM. For real this time

charkour:
Thanks for the PR! I appreciate the code comments.

charkour (changes requested):
Looks good. I'd recommend writing some tests using jest in `facultyLoadsService.spec.ts`

kvlinden:
Ok, I added the unit tests and double-checked that the popups are still functional.

charkour (approved):
Tests look good thanks!

PR #173, opened 2021-03-30
Hide classes spanning more than one day

### LINKED ISSUE #160: Schedule disappears if a section starts late and ends after midnight
I'd assume there wouldn't be any sections that intentionally get entered at this time, but I tried to update a 12:30 class to 11:30 but forgot to switch it to AM and the entire schedule disappeared and was unrecoverable without reimporting.

bryantgeorge:
To Test:
- Add sections/change existing sections to span more than 1 day (ex: running from 11:30PM to 12:20AM)
- Try importing a schedule with such a section
- These sections should no longer cause the schedule to disappear because these sections will no longer be displayed on the faculty/room schedules (but they are output and displayed on the teaching loads tab)
- May want to test edge cases like sections spanning more than 2 days (ex: starting at 11:30PM and going for 1500 minutes)

Note that I think only the changes in `getMinAndMaxTimes()` are necessary to solve the problem, but I kept the changes in `getEvents()` to be safe.

Closes #160

### DIFF
--- a/client-course-schedulizer/src/utilities/services/scheduleService.ts
+++ b/client-course-schedulizer/src/utilities/services/scheduleService.ts
@@ -46,33 +46,36 @@ export const getEvents = (schedule: Schedule, groups: "faculty" | "room"): Group
           const group = groups === "faculty" ? prof : room;
           const startTimeMoment = moment(meeting.startTime, "h:mm A");
           const endTimeMoment = moment(startTimeMoment).add(meeting.duration, "minutes");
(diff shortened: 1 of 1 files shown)


charkour (approved):
This is great! Thanks for the code comments. Works as expected. 

I'm glad I can edit those weird classes on the teaching loads tab and not have it break the schedulizer tab. Having code make sure it's safe in two places is good. Thanks for the fix.

PR #176, opened 2021-04-01
Connect Harmony with Schedulizer

charkour:
In this PR:

- creates a global results store for harmony results
- converts the harmony output to our schedulizer format
- displays harmony output on a visual schedule
- displays a button to send the created schedule to the schedulizer
- separates schedule components to separate files
- move helper functions below components in same file
- rename props to be `ComponentNameProps` rather than a shadow of `ComponentName`. This allows for easier imports without name conflicts
- adds `readonly` prop to `Schedule` component that does not do anything yet
- add `useAppContext` hook
- add `useRedirect` hook

To test:
- go to harmony page
- add two professors
- add four courses with the format prefix-number-letter
- add two rooms
- add times with the format xx:xx
- click submit on all of the four forms
- then select all checkboxes for the assignments
- then click `find schedule` and a schedule should be found
- then click `send to schedulizer` and the new schedule should show up on the schedulizer page

Right now, Harmony is quite cumbersome to use, do you have suggestions on how to fix that? Maybe a select all checkboxes button and separating the different parts to their own pages with a `PageStepper` component?

Thanks for the comments and critiques!

closes #136

PR #177, opened 2021-04-01
Improvement/reduce scrolling

### LINKED ISSUE #88: Reduce Amount of Scrolling
There is a lot of scrolling require to see everything currently.

Reduce screen real-estate used and/or create another tab focused on visualizing at a wider view.

**Additional context**
See Pruim's original schedulizer for potential inspiration.

bryantgeorge:
This PR generally tries to make less scrolling required, to do so it:
1. Makes widths (and text sizes) of faculty/room schedules constant pixel sizes, so that in theory you should be able to fit 5 schedules instead of 4 on a page with a sufficiently wide monitor. Notice that when expanding the width of the browser window, the schedules stay the same width and more schedules are available.
2. Creates a `Department Schedule` tab which has the courses grouped by the faculty name (should be the same for all courses in the math schedule). These schedules are 4 times as wide, so that you can see the courses. This is intended to help dept chairs to see the entire schedule at a glance.

Things I would recommend testing:
1. Try on chrome, firefox, and safari (if possible).
2. You could try the department view for the `Course Section Enrollment.csv` which looks really cool to have all of the department schedules in a semi-efficient space. (In my experience, using this schedule can be really computationally intensive and switching from the `Faculty Schedule` tab to the `Department Schedule` tab will take a while.)

Closes #88

### DIFF
--- a/client-course-schedulizer/src/utilities/hooks/addSectionHooks.ts
+++ b/client-course-schedulizer/src/utilities/hooks/addSectionHooks.ts
@@ -118,7 +118,7 @@ const scrollToUpdatedSection = (newCourse: Course, newSection: Section) => {
     });
   });
   if (newElement && newElement.parentElement?.parentElement) {
-    newElement.scrollIntoView({ behavior: "smooth", block: "nearest", inline: "nearest" });
+    newElement.scrollIntoView(false);
   }
 };
 
--- a/client-course-schedulizer/src/components/Tabs/DepartmentSchedule/DepartmentSchedule.scss
+++ b/client-course-schedulizer/src/components/Tabs/DepartmentSchedule/DepartmentSchedule.scss
@@ -1,0 +1,3 @@
+.department-calendar-width {
+  width: 1400px !important;
+}
--- a/client-course-schedulizer/src/components/Tabs/DepartmentSchedule/DepartmentSchedule.tsx
+++ b/client-course-schedulizer/src/components/Tabs/DepartmentSchedule/DepartmentSchedule.tsx
@@ -16,6 +16,7 @@ export const DepartmentSchedule = () => {
       <Schedule
         calendarHeaders={departments.sort()}
         groupedEvents={getEvents(schedule, "department")}
+        scheduleType="department"
       />
     </>
   );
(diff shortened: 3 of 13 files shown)


charkour:
In #176, I split all of the Schedule.tsx components into their own files so there will be some conflicts with this PR. Once this PR is approved, we can marge this in and I'll worry about reconciling the issues! 

I'll check this out tomorrow likely! Thanks for the good work.

client-course-schedulizer/src/utilities/hooks/addSectionHooks.ts
```diff
@@ -118,7 +118,7 @@ const scrollToUpdatedSection = (newCourse: Course, newSection: Section) => {
     });
   });
   if (newElement && newElement.parentElement?.parentElement) {
-    newElement.scrollIntoView({ behavior: "smooth", block: "nearest", inline: "nearest" });
+    newElement.scrollIntoView(false);
```
charkour, line 121: Is this change on purpose?

charkour (changes requested):
Looks good to me! Thanks for updating the styles and the dept tab looks really cool! If we had more time, we could look into memorization and virtualization to increase rendering performance, but tbh I don't understand either of those technologies. 

Nice work.

client-course-schedulizer/src/utilities/hooks/addSectionHooks.ts
bryantgeorge: Yes, I forgot that I did this, but this is to make it compatible with Safari, see:

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView
image

Safari doesn't support the `scrollIntoViewOptions` parameter, but they do support the `alignToTop` boolean parameter. Unfortunately, this will cause the scroll into view to not be smooth, but I think it's worth it. (Also the scroll into view for the faculty schedule still has the `scrollIntoViewOptions` parameter, but I figure if that one doesn't work on Safari it isn't the biggest deal)

charkour: Gotcha, thanks!

PR #178, opened 2021-04-03
Allow users to delete sections without meetings

### LINKED ISSUE #170: Cannot Delete Section with No Meetings
If a section does not have a meeting, it cannot be deleted. To reproduce, add a section but do not include a duration. Then, go to the faculty loads tab and attempt to delete that section. It will not be deleted.

bryantgeorge:
This PR has some pretty fundamental changes again to how meeting/section/course updates/adds/deletes are handled, so it might be best to have at least 2 reviews.

To Test:
- Can reproduce by "add a section but do not include a duration. Then, go to the faculty loads tab and attempt to delete that section" according to #170 
- If there are duplicate meetings, deleting will now no longer delete all of them (instead it will just delete one)
- This PR also ensures that if you click `Add New Section` on an existing section while adding a new meeting, any changes made to the course and section will also be applied. For example, if you have `MATH-100-A` which meets `MWF at 8:00`, and you add a new meeting `TTh at 10:30` but you also change the `Faculty Hours` and `Department` fields, then a new meeting will be added, the `MATH-100-A` section will be updated to have the new `Faculty Hours` value, and the `MATH-100` course will be updated to the new `Department`. (Before, the course stuff would get updated properly, but the section stuff wouldn't).
- This PR also ensures that if you update a section and/or course and change identifying info about it (like the prefixes or instructors for example), it will update that info for that section/course. (See kvlinden's comment in #170 about an example of this that should be fixed here)
- May also want to try other things related to adding/deleting sections/meetings to make sure this doesn't break anything else

One thing I noticed while working on this is that if you have a prof who teaches multiple meetings of a given section, only one of them is listed when accessing it from the teaching loads tab (and deleting it from the faculty loads tab only deletes the displayed meeting). Is this the intended behavior?

Closes #170


Jdellis818:
@bryantgeorge could you resolve the merge conflicts before we review? That way we can make sure that nothing gets messed up while resolving them.

bryantgeorge:
> @bryantgeorge could you resolve the merge conflicts before we review? That way we can make sure that nothing gets messed up while resolving them.

Oops my bad, for some reason I didn't see that (do merge conflicts not show up until after you unmark as draft?), in any case, it should be good now.

Jdellis818 (approved):
LGTM! This is a great fix! One thing I will note is that updating a course number will update it for all sections of that course. This could be what the user would want, I'm not sure exactly, but I think sometimes they may enter all the information for a section but get the number wrong, update just the number, and be surprised when all of the sections for that course are changed too. This is probably more of an overall system issue than with this PR.

PR #180, opened 2021-04-13
Check to make sure the meetings are not the same section for conflicts

Jdellis818:
There was a bug with the conflicts where 2 meetings for the same section would appear as a conflict. This PR adds a check to make sure the two meetings are not for the same section before marking it as a conflict. It also adds a utility function to get a string of the location (building and room number separated by a space).
To test:
- Go to develop and load this csv: https://raw.githubusercontent.com/kvlinden/kvlinden.github.io/master/data/full_schedule_2021-04-07_submitted.csv
- Notice that there are sections with red borders that aren't conflicts, such as DATA-202 (they're marked as a conflict because they have 2 rooms)
- Now, load the same csv on this branch and the faulty conflicts should be gone


client-course-schedulizer/src/utilities/services/conflictsService.ts
```diff
@@ -55,7 +57,9 @@ export const findConflicts = (schedule: Schedule): Schedule => {
         range1.overlaps(range2) &&
         meeting1.term === meeting2.term &&
         meeting1.days.some(meeting2IncludesDay) &&
-        (meeting1.instructors.some(meeting2IncludesInstructor) || meeting1.room === meeting2.room)
+        (meeting1.instructors.some(meeting2IncludesInstructor) ||
+          meeting1.room === meeting2.room) &&
+        meeting1.sectionName !== meeting2.sectionName
```
bryantgeorge, line 62: I think this is good because in theory, this should be correct 99.9% of the time, but I just wanted to point out as a note that `getSectionName()` just uses the first prefix, so there is a chance that it mistakenly thinks 2 sections are the same (but this would require a bad schedule where there was something silly like `MATH/PHIL-100-A` and `MATH/CS-100-A`). Again, this is just a note and probably doesn't require a change.

bryantgeorge (approved):
LGTM, just resolve the conflict and this should be good (I really like the change saying that the conflicts tab isn't supported, that's probably more clear than Item Four)

client-course-schedulizer/src/utilities/services/conflictsService.ts
Jdellis818: Ah yeah that is true. I didn't think about that. Good catch!

PR #195, opened 2021-05-04
add ability to view a demo schedule

charkour:
This adds the ability to click a "View Demo" button for a user to load a schedule if they don't have one.

It will be helpful for people going to our website tomorrow so they can try out the app.


Jdellis818 (changes requested):
This is great! The only change I'd suggest is to use the CS schedule. We can either add it to our csv folder or use this link: https://kvlinden.github.io/data/full_schedule_2021-04-07_submitted.csv

charkour:
Good idea, thanks!

Jdellis818 (approved):
LGTM!

Issue #256, opened 2022-11-14
Replace node-sass with sass?


rpruim:
See https://sass-lang.com/blog/libsass-is-deprecated:

> After much discussion among the Sass core team, we’ve come to the conclusion that it’s time to officially declare that LibSass and the packages built on top of it, including Node Sass, are deprecated. For several years now, it’s been clear that there’s simply not enough engineering bandwidth behind LibSass to keep it up-to-date with the latest developments in the Sass language (for example, the most recent new language feature was added in [November 2018](https://github.com/sass/libsass/releases/tag/3.5.5)). As much as we’ve hoped to see this pattern turn around, even the excellent work of long-time LibSass contributors Michael Mifsud and Marcel Greter couldn’t keep up with the fast pace of language development in both CSS and Sass.

> **If you’re a user of Node Sass, migrating to Dart Sass is straightforward: just replace node-sass in your package.json file with sass. Both packages expose the same JavaScript API.**

charkour:
I'd recommend switching from Sass to [TailwindCSS](https://tailwindcss.com/). Long story short, tailwind makes managing styles for component-based frameworks (like React) a breeze. Sass is an older CSS technology and development is limited. Also, tailwind is much more efficient than Sass in terms of bytes sent across the network.

rpruim:
I've tested the switch to sass on my mac and after changing one line in package.json, I'm now able to build the project.  So we should 
1. Make that change ASAP since it appears to be (a) non-breaking, and (b) required at least on some platforms.
2. Consider making the change to TailwindCSS.  But this is likely less urgent.

rpruim:
Here's what I put in package.json in place of node-sass:

```
   "sass": "1.56.1",
```

charkour:
Looks good to me!

PR #257, opened 2022-11-14
migrate from node-sass to sass

rpruim:
See issue #256.

This implements a one-line fix to migrated from the deprecated node-sass to (dart) sass.


charkour (approved):

rpruim:
this is subsumed by #258.

rpruim:
Since #258 has been accepted, we don't need this one.

Issue #250, opened 2022-10-27
Reorder the columns of the CSV output


rpruim:
The schedules will need to be hand entered into Workday.  Workday has a long form for this, and it will greatly speed up the process for data entry if we can put our csv columns in an order that mirrors the data entry form.  Here is what Melisa Hubka suggests:

> Here’s the list of columns in the preferred order:
>  
> Section name: MATH 171 A
> Start/end dates
> Title
> Fac Load
> Section Capacity
> Sem Hours
> Notes to be shown to stu/faculty (if there are any)– this requires clicking on the section from Find Course Sections
> Delivery mode: in person, online, hybrid (rare)
> Instructor(s)
> Meeting Pattern – which is days then time
> 
> Instructional format defaults from the course (lecture, lab, etc) so we don’t need that information – or it could be to the far right.

rpruim:
Raymond has requested another reordering:

```
Term
Section name: MATH 171 A
Start/end dates
Title
Fac Load
Sem Hours
Notes to be shown to stu/faculty (if there are any)– this requires clicking on the section from Find Course Sections
Delivery mode: in person, online (syn/async), hybrid (rare)
Instructor(s)
Meeting Pattern – which is days then time
Room
```

kvlinden:
The "Teaching Loads" tab - "Other Duties" column get messed up in the production branch now and I suspect that this column-output selection issue is the change that messed things up. To reproduce the error:

1. Read [this 2023 CS schedule](https://senior-knights.github.io/course-schedulizer/#/?csv=https://kvlinden.github.io/data/full_schedule_2023.csv)
2. Go to the "Teaching Loads" tab and note the stuff in the "Other Duties" and "Other Hours" columns.
3. Write it out and read it back in, and then revisit the other duties/hours.

The import code appears to be keying off of the "SectionName" column (in a strange manner). Adding that field to the CSV and setting it to "--" signals an other duties/hours entry, using the InstructionalMethod column as the other duty name. But note that when that column is added, all the section names for all the courses must be included, otherwise they all get moved to the other duties/hours column.

The workaround is painful: emport the CSV and then manually add the SectionName column with the appropriate values, i.e., either "--" or =CONCATENATE(C17,"-",D17,"-",E17)

rpruim:
Ack..  I'm guessing there was no test to alert me to this and I didn't catch the non-teaching duties bit.

[...]

PR #267, opened 2022-11-18
reorder columns in exported CSV. remove unused variables.

rpruim:
This takes care of issue #250 and eliminates some of the warnings that had been generated previously.


rpruim:
This is subsumed by #269, so I'm closing.

PR #269, opened 2022-11-18
Feature/issue268 convert instructional method

rpruim:
This PR just maps things like LEC, SEM, and LAB to 'In-person', since that will most likely be the delivery mode for those types of courses.

This is something that will likely only be needed for the transition year as we migrate from Colleague to Workday.

### DIFF
--- a/client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
+++ b/client-course-schedulizer/src/utilities/helpers/caseFunctions.ts
@@ -196,7 +196,7 @@ export const timestampCallback = (value: string, { section }: CaseCallbackParams
 };
 
 export const instructionalMethodCallback = (value: string, { section }: CaseCallbackParams) => {
-  section.instructionalMethod = value;
+  section.instructionalMethod = instructionalMethod2deliveryMode (value);
 };
 
 export const sectionCallback = (value: string, params: CaseCallbackParams) => {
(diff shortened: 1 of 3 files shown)


rpruim:
@kvlinden, I think this another PR that ww should accept.  But read on before agreeing.

It is a bit of a hack to bridge between the two systems.  The old system had "Instructional Method", the new has "Delivery Mode".  This PR simply converts some of the old labels to "In-person" to help correctly seed the data.

Another option would be to ignore the "Instructional Method" altogether and add a column correctly named "Delivery Mode".  For this year, people would have to enter that data for every section (unless we decide to have "In-person" be the default, which could lead to some errors). 

I might do a little exploring to see whether my alternative approach is quickly doable.  I'm understanding the data setup and import/export process better now than I was a week ago, so it might be doable.

Thoughts?

rpruim:
Well,  I did some digging.  In a local branch I've added a DeliveryMode column.  This required editing 10 different files.

For now, I'm leaving InstructionalMethod as is.  This field is also used to record non-teaching activities.  My thinking is that we could perhaps rename this later to indicate that use, and stop using it to track the no-longer-needed InstructionalMethod.

For now, I'm NOT auto-populating DeliveryMode.  This will mean that creators of schedules will need to go through an edit that field for every course (in Excel or in the app).  That is a bit of a nuisance, but probably less likely to introduce errors due to oversight.

I'll do a bit more testing a little later.  If things look good, we can probably disregard this PR in favor of a new one I will create.

PR #278, opened 2022-11-22
Half semester conflict

rpruim:
I improved the test for term overlaps and added low-level tests to confirm that it is working as intended.

### DIFF
--- a/client-course-schedulizer/src/utilities/services/conflictServices.spec.ts
+++ b/client-course-schedulizer/src/utilities/services/conflictServices.spec.ts
@@ -1,0 +1,37 @@
+import { sign, rangesOverlap, numericalSemLength, termsOverlap } from "./conflictsService";
+
+[
+    {in: -2, out: -1}, 
+    {in: 3, out: 1}, 
+    {in: 0, out: 0}
+].forEach((test_case) => {
+    test("sign function works correctly", () => {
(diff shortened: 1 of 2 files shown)


rpruim:
@kvlinden, @charkour -- either one of you want to review this?  It would be nice to get this into production if it isn't causing any problems.

charkour:
I can take a look. 

Would you be willing to email me a schedule I can test with, @rpruim?

rpruim:
I emailed some example files.  There is also a link in #250 to an example from CS.

rpruim:
I think there are some older examples in the repo as well (perhaps used for some of the tests?).  I haven't dug into that too deeply yet, but we should probably create some new import/export tests since we have chanced the import/export process.

charkour (approved):
Code looks good. Thanks for the unit tests!

rpruim:
I just deployed this.

rpruim:
@charkour, thanks for approving.

@kvlinden, you could load your CS schedule and see if there are any remaining non-conflicts being flagged as conflicts.  I didn't immediately spot any, but you can probably go through the list more quickly than I can since I need to open up the sections to see when they are scheduled and you might just remember.

kvlinden:
The current production branch appears to be working properly. Thanks!

PR #289, opened 2022-11-25
Bugfix/section name for other duties

rpruim:
This PR addresses another case where non-teaching status needed to be inferred (when adding a new course rather than a new section).  See #250 for discussion.

### DIFF
--- a/client-course-schedulizer/src/utilities/services/addSectionService.ts
+++ b/client-course-schedulizer/src/utilities/services/addSectionService.ts
@@ -93,7 +93,13 @@ const convertToSemesterLength = (sl: Half | Intensive | SemesterLengthOption): S
 };
 
 export const getSectionName = (course: Course, section: Section) => {
-  return `${course.prefixes[0]}-${course.number}-${section.letter}`;
+  return `${course.prefixes.length ? course.prefixes[0] : ""}-${course.number}-${section.letter}`;
+};
+
+// if isNonTeaching hasn't already been set, infer it from the 
+// section name (computed from prefix, course number, and section letter all being empty)
+export const isNonTeaching = (course: Course, section: Section) => {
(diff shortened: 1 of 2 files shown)


client-course-schedulizer/src/utilities/services/addSectionService.ts
```diff
@@ -93,7 +93,13 @@ const convertToSemesterLength = (sl: Half | Intensive | SemesterLengthOption): S
...
 
 export const getSectionName = (course: Course, section: Section) => {
-  return `${course.prefixes[0]}-${course.number}-${section.letter}`;
+  return `${course.prefixes.length ? course.prefixes[0] : ""}-${course.number}-${section.letter}`;
+};
+
+// if isNonTeaching hasn't already been set, infer it from the 
+// section name (computed from prefix, course number, and section letter all being empty)
+export const isNonTeaching = (course: Course, section: Section) => {
+  return section.isNonTeaching || getSectionName(course, section) === "--";
```
charkour, line 102: All I would say here is to pull this string out into a variable to give is a name. Right now, this is a "magic string" so it's a little hard to know why `--` is special. Or you could link to the GitHub discussion. 

`const NON_TEACHING_CLASS_TITLE = "--";`

charkour (approved):
Looks good. Just have one suggestion.

client-course-schedulizer/src/utilities/services/addSectionService.ts
rpruim: Looks like a good amendment.

rpruim: The comment was supposed to indicate why the right side of || is being computed, but we could be even more explicit.

Issue #268, opened 2022-11-18
Change LEC, SEM, etc to things like In-person


rpruim:
This is a shift from the old system to the new, and when people import a schedule, they will get the old-style data here.  It would be good to auto-convert these to save chairs the pain of hand editing every one of these, or the registrar's office the hassle of getting the wrong sort of data.

rpruim:
PR #269 closes this.

PR #293, opened 2022-11-26
Feature/delivery mode

rpruim:
I've added a columns to the data: delivery mode.

* This will be the new lingo going forward in the workday world
* I've left instructional method in place because
    * it is used to track non-teach activities (so we could just rename it and use it for that purpose ONLY)
    * technically, instructional method still exists, it just isn't something that people creating the schedule can change on a per-section basis -- it is already in the workday system at the course-level (ie, decided upon when the course is proposed, not when the schedule is created).

The behavior of the pull request:

* uses the add section popover to edit the delivery mode field
* uses the instructional mode field for non-teaching activities
* leaves any data in delivery mode as is from old schedules (so you will see things like LEC, SEM, etc. for now).  

I've done several tests of creating classes and not teaching activities, importing and exporting.  So far, everything I've tried has worked as I've expected.

This deals with #268 and supersedes #269.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,44 +26,45 @@ For more information, please visit the [About Page](https://senior-knights.githu
 
 ## How to use
 
-Detailed information can be found on our [Help Page](https://senior-knights.github.io/course-schedulizer/#/help) (once we make it). Access our production website and upload a CSV following the prescribed specifications:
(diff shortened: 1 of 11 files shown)


kvlinden (approved):
LGTM

PR #294, opened 2022-11-26
Update deploy.yml to use Node v14

charkour:
Node v12 is [deprecated](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) as a possible GitHub Actions runner.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -11,7 +11,7 @@ jobs:
       working-directory: ./client-course-schedulizer
     strategy:
       matrix:
-        node-version: [12.x]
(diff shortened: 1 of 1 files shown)


charkour:
@rpruim, this should resolve the GitHub action deployment issue!

rpruim (approved):
Thanks.

rpruim:
@charkour , I'm not sure we have completely solved the problem.  I got another failure notice after the upgrade to v 14.

charkour:
Hmm okay I can look I to it.

PR #296, opened 2022-11-27
Feature/standard times

rpruim:
I have an implementation of standard time checking.  Here are some features:

* the legal times are stored in times.json (in src/data)
* nonstandard times lead to orange borders 
* no text in the conflicts table, but we could add that

I worked off a preliminary version of the master schedule, and I see now that not only are the time slots prescribed, but actual meeting patterns are prescribed (some days @ some time for some duration).  My check would allow any course to meet only some of of the prescribed days, or if MWF and TR were allowed, you could schedule things TWR -- I'm only checking that a legal day/start/duration combo was selected for each session, not that the entire meeting pattern is legit.

I think this is sufficient for now, but it shouldn't be too hard to switch over to the picker test.  In some ways, the logic might actually be a bit easier.  But we might want to be a bit more flexible and allow subsets of meeting patterns. I think there will be a number of things, like seminars, that meet just one of the days in a given meeting pattern.  I'll give this some more thought.  I might even try changing this yet tonight.

Things are reasonably well encapsulated, so we would just need to modify `isStandardTime()` (and possible the format of the data holding the standard times).  The excel spreadsheet from the registrar (second tab) could form that basis of full meeting pattern data set.


rpruim:
I've updated this to implement the strictest check.

TODO: Consider implementing the "subset of a legal meeting pattern" check.

kvlinden (approved):
LGTM

Issue #300, opened 2022-11-28
modify check for non-standard times


rpruim:
In conversation with @kvlinden, I think we have come up with a middle-road for checking standard times.

Given a meeting patterns with several days, a start time, and a duration, we will allow

* a course to meet all of those days at the stated time and duration
* a course to meet exactly ONE of those days at the stated time and duration.

This will avoid flagging things like seminars that use just one slot (and help people get those scheduled correctly).

I think I have an approach that will make this fairly easy to implement.  Basically, I'll add additional meetings patterns that include just one day as well as the meeting patterns that use all the days when I build the data that is checked against.

rpruim:
Closing as completed.

Issue #295, opened 2022-11-26
Convert from TH to R


rpruim:
My understanding is that we will be moving toward using R as the standard abbreviation for Thursday.

We probably want to read TH or R but write R.

PR #301, opened 2022-11-28
Feature/standard times issue 300

rpruim:
As per #300, I've made the standard time check a bit looser.  This will allow people to schedule classes that meet only one time from a meeting pattern without being flagged.

I've implemented this be expanding the list of legal meeting patterns when it is created to include the one-day versions of teach meeting pattern.  Everything else is just as before.

I've added several additional test cases to confirm the new behavior is working as desired.

Side note: See #295.  This would be cleaner if I didn't need to convert from TH to R and back again.


kvlinden (approved):
LGTM

rpruim:
I'm confused about this PR.  It appears the changes it is intended to introduce are already in production.  Furthermore, there is a conflict with a subsequent update to the standard meeting patterns. I've twice tried to resolve those on github, but they don't seem to take.

I'm going to close this since I don't think it is actually needed.

charkour:
> It appears the changes it is intended to introduce are already in production.

Did you push to production on accident at one point? Or were they included with a different feature branch? It seems the latter is most likely (I've done that before).

rpruim:
I don't think I can push to production.  (I tried it once to see if that's what happened and my push was blocked.)

I'm not sure just what happened, but if you are interested, this is roughly how things went down (IIRC):

* I used a branch to make a pull request, then I continued to work in that branch so I could update the PR.
* Meanwhile @kvlinden approved the PR (and I may have merged it at that point, but I don't remember exactly).
* Github cleans up old branches involved in PRs, so when I tried to push to my branch, it was longer on the remote.
* So I made a new branch at that commit and pushed that and issued a PR that had mostly the same commits, but another added on.
* Meanwhile, we approved and merged a different PR to fix the data in the meeting patterns, that caused the this second PR to have conflicts.
* I tried resolving them on github (because I wanted to see how that worked), and I clicked on resolved, but the resolution didn't seem to persist.  In fact, I did that twice.

charkour:
Ah yeah, sounds like a complicated git workflow. I wish GitHub had the option to disallow fixing conflicts online. Often times, it doesn't work quite right.

PR #302, opened 2022-11-29
Bugfix/meeting patterns data

rpruim:
I had Todd Dornbos look over meeting-patterns.csv and we discovered an error.  (Some data that I had edited by hand was off by a few rows and I also saw that what he sent me had H instead of R for a Thursday in one spot.)

The only changes in this PR

* updated CSV
* updated JSON
* added short R script that converts from CSV to JSON.

We could probably just ingest the CSV in the app, or convert to JSON without using R, but this was the quick and easy way I knew how to do it.


kvlinden (approved):
LGTM

Issue #308, opened 2022-12-02
Feature request: Add toggle for full term, 1st half, 2nd half, etc.


rpruim:
This might be easier than trying to create a single display that shows first and second half courses (and intensives) in some common view (see #244).  It basically amounts to adding a filter on the data before generating the views.

I suggest doing this first and then revisiting the other.

I think that we want to create the filter to show anything that overlaps with the selected time period.  So if someone selects first half semester, they should also see the full term courses, since those also meet in the first half.

We can probably recycle some of the code from the conflict detection for partial-semester courses.

PR #315, opened 2022-12-09
add selector for semester part

rpruim:
This addresses #308.

I've added a selector for semester part.  It works much like the color-by selector.  All courses with intersecting terms are displayed. So if you pick "First Half", you will see Full, First Half, A, and B courses since all of those meet during the first half of the semester.

This my need revisiting if/when we implement custom terms.  But that is disabled for now.


rpruim:
I deleted the search bar from the UI since it is disabled and restyled a bit given the new selector.

kvlinden (approved):
LGTM

Issue #305, opened 2022-12-02
Feature request: Add customizable "group" field that can be used to colorize schedule


rpruim:
This came up several times at the workshop.  Example use cases:

* Flag courses by levels other than 100/200/300 level (nursing)
* Flag core courses or tagged courses (several)

This seems both relatively easy and pretty useful, so it seems like a useful thing to implement.

rpruim:
Just dropping in an encouraging testimonial:

> Randy,
 
> I really appreciate the addition of the “Group” field. Being able to display upper level courses with different colors by concentration is a big help for me!

PR #316, opened 2022-12-10
Feature/add group

rpruim:
I've added a group entry field in the popover and the ability to color by these groups.  See #305.

Note: This PR includes the changes from #315. So best to approve that one first.


kvlinden (approved):
LGTM

PR #318, opened 2022-12-12
add selectedSemesterPart dependency to useMemo()

rpruim:
Fixing a warning from npm run build

I'm also leaving a warning about not using the search bar -- I expect that to return.


charkour (approved):
Generally, you should always follow the ESLint rule and all variables accessed inside the callback to the dependency array. It's rare to omit one.

rpruim:
> Generally, you should always follow the ESLint rule and all variables accessed inside the callback to the dependency array. It's rare to omit one.

I think it must have been an accidental "undo" on my part.  I know that I had that there at one point and didn't intend to remove it.  But I didn't catch it until I was about to deploy.

charkour:
Roger, gotcha!

Issue #90, opened 2021-02-08
Implement Search


bryantgeorge:

kvlinden:
We can use the standard browser search in place of this for some things.

rpruim:
I played around with this a bit today, and modified the search to be a filter:

![image](https://user-images.githubusercontent.com/722231/207158295-56790f80-1abb-4b68-8f7b-ce14d8f2bcd2.png)

It is mostly working, but there seems to be a bug when the filter bar is cleared (at least sometimes) that leads to

```
 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, 
but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```
and causes the app to freeze.

I'll have to look into this more later.  But if @charkour has any ideas, feel free to share.

charkour:
Ah yeah, those are some messy bugs. I could take a look later this week. 

Do you have a branch with this change?

rpruim:
The searching/filtering is a regex match on a stringified version of most of the information maintained about a section (location is intentionally left off to make course number searching work better, for example).

rpruim:
The branch is local, but I can push it.

rpruim:
https://github.com/senior-knights/course-schedulizer/tree/feature/implement-search-filter

I did this mostly by imitating some other things (and small amount of documentation reading), which is a little bit dangerous. I didn't interact directly with setState() or useEffect(), but I did use useMemo().

So far the most reliable way to get the error is to clear the search bar, but it doesn't happen every time, and I don't promise there are not other ways to trigger it as well.

rpruim:
In case it is useful, here is a longer bit of the error message:

```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    in Stick (at ScheduleBase.tsx:90)
    in div (at ScheduleBase.tsx:89)
    in div (at ScheduleBase.tsx:87)
    in ScheduleBase (at Schedule.tsx:17)
    in div (at makeFC.tsx:26)
    in Loaded (at Schedule.tsx:16)
    in AsyncComponent (at Schedule.tsx:14)
    in Schedule (at DepartmentSchedule.tsx:16)
    in DepartmentSchedule (at Tabs.tsx:63)
    in div (created by ForwardRef(Typography))
[...]

PR #323, opened 2023-01-04
Feature/update meeting patterns

rpruim:
From Todd Dornbos:

Update – After a conversation with Arlene, we have decided to alter one of the meeting patterns.
 
Row 45 on the ‘Meetings Patterns’ sheet – This 3x120 min option has changed to 8:20am-10:20am (previously 8:00am to 10:00am).

I've updated the CSV and JSON files to reflect this change from the provost's office.


rpruim:
Hmm.  I'm going to close this.  It looks like this grabbed a commit I don't think we want since the search bar sometimes leads to things crashing. (See #90 for discussion.)

I'll redo this to limit the changes and submit a new PR.

PR #339, opened 2023-11-16
move start times from 18:30 to 19:00 for most evening slots

rpruim:
This simply edits the files containing start times of courses to reflect a change for the 24-25 academic year.

Note: A fancier version of this would take the academic year into account so that courses are not flagged in old schedules.  That would require handling academic year more explicitly (which might be a good idea).


kvlinden:
We're going to merge into develop first, to make that everything is working.

kvlinden:
We're going to try to merge your branch into develop first, to make sure that we can deploy it all properly. The errors you're seeing here were, we think, fixed in both fix-local and develop.

We thought about modifying your branch but decided to be safe and leave it as it is.

kvlinden:
@rpruim - We now have your changes plus the fixes we've made running on the develop branch, but it's not signalling a constraint violation on 6:30pm courses. Are you certain the changes you made fix the issue.

kvlinden:
@rpruim - I'm cleaning up some old Schedulizer branches/PRs and note that you've got this branch/PR and three other feature/ branches out there that are 2-3 years old. Are they still relevant?

kvlinden:
This one is really old, and the 7pm times are fully set in the system.

PR #348, opened 2024-02-28
Avoid caching remote schedule data. Add only this element to develop.

NycoCC1993:
Docs: https://developer.mozilla.org/en-US/docs/Web/API/Request/cache


kvlinden:
Did you test this locally? 
Once tested, we'll need to re-deploy the whole system to the development server, not just the one file.

NycoCC1993:
> Did you test this locally? Once tested, we'll need to re-deploy the whole system to the development server, not just the one file.

Applying the change locally resulted in no errors or issues that I could detect.

kvlinden (approved):

PR #358, opened 2024-07-26
Fixed versioning issues

kvlinden:
This release enforces specific versions of Node and pnpm, and fixes some npm package version issues.
We will still need to review/update the package versions, but this set is consistent and runs properly.


kvlinden:
going to push this config update unilaterally as well

Issue #372, opened 2025-03-12
Missing export columns (develop branch)


rpruim:
I've been doing some schedule stuff with my 2025-26 schedule and had been counting on using the time stamp and comments columns to track the changes to give to the registrar's office.  Trouble is -- neither of these is populated in my exported Excel file.

It does export to the CSV, so I have the data recorded there.

Merging with develop and restarting did not seem to fix the issues.  Is this just an oversight in migrating to the new format?

PR #374, opened 2025-04-02
Feature/delivery mode interface

wahrheit319:
this should fix missing columns, including comments


wahrheit319:
fixed issue #372

kvlinden:
Is that the field order that they want? They used to put delivery mode just before the comment, not after it. 
I'd suggest that change.

kvlinden (approved):
LGTM

Issue #368, opened 2025-02-24
Add ability to toggle which subset of multiple schedules is displayed


rpruim:
If we implement #366, it would be good to be able to include any subset of the schedules in the views.

Issue #367, opened 2025-02-24
Colorize exported schedule based on differences to a reference schedule


rpruim:
Once we can load two schedules (see #366), one use case would be to export a version of one schedule with cells highlighted if they differ from a reference schedule.  This would be handing for

a. Checking that the submitted schedule matches what gets entered into the system
b. Highlighting changes when submitting amendments
c. Highlighting year over year changes

Different use cases might have different definitions of what makes a row the same or different. A simple interface might be checkbox group to select columns.  Any rows that are identical in those columns would be considered the same.  Since there may be multiple rows in a single schedule that match on a set of columns, we should also compute a tally of rows or faculty load as part of this process.  This could lead to the marked data having fewer rows than the full schedule.  (This would be akin to applying a groupby on the checked columns and then adding some summary columns for number of sections, total faculty load, etc.)

The highlighted version (which may have fewer rows) should be exported as a separate file or on a separate sheet.

Issue #366, opened 2025-02-24
Implement ability to load multiple schedules at once


rpruim:
This could have a number of use cases. See separate issues for use cases.

Related issues:
* #367
* #368

PR #376, opened 2025-04-04
Feature/compare

wahrheit319:
Implemented issues #366, #367, and #368.
Implemented a meta tab that reads/writes notes, version, academic year, and date.

known problems:
1. Unable to export Excel with colors.
2. If two compared schedule file names are the same, the comparison export will crash.

notes:
1. The exported grouped sheet format might be completely incorrect, but I'm not sure.
2. When you export to Excel, it will export all the schedules you have toggled into one schedule. You can combine multiple schedules if you want.
3. When multiple schedules are toggled simultaneously, any section updates will apply to all toggled schedules. This is an intentional feature that allows batch editing across selected schedules. When only a single schedule is toggled, changes will be confined to that specific schedule only.
4. Modify "Reset to Default" columns in ScheduleComparer.tsx, line 840.
5. This may be completely different from what you asked for, and it might be a total piece of hot garbage, but I have tested it with different types of files, and I believe it works well, and I had a fun time doing it.


kvlinden:
We'll need to remember to merge [the URL argument fix commit](https://github.com/senior-knights/course-schedulizer/pull/385/commits/b7ee7d9c259d162ccce1ee10d17ef22bf69eb59d) into develop also.

kvlinden (approved):
LGTM

But, there are some conflicts that need to be resolved. Any idea what's happening there?

PR #378, opened 2025-04-14
Update/help

kvlinden:
Updated the help text to include saving a schedule and to be compatible with the new XLSX file handing feature.


kvlinden:
Oops, I wanted this to merge into develop.

PR #379, opened 2025-04-14
Updating the help for the new XLSX format features

kvlinden:
Updated the help text to include saving a schedule and to be compatible with the new XLSX file handing feature.
This replaces my original PR what incorrectly merged into production.


wahrheit319 (approved):
It works

PR #381, opened 2025-04-15
updated schedule editing help instructions

kvlinden:
A couple more edits to the help instructions that explain how to get a schedule file to edit

### DIFF
--- a/client-course-schedulizer/src/components/Tabs/tabComponents/NoCoursesHeader/NoCoursesHeader.tsx
+++ b/client-course-schedulizer/src/components/Tabs/tabComponents/NoCoursesHeader/NoCoursesHeader.tsx
@@ -9,7 +9,7 @@ export const NoCoursesHeader = () => {
     <>
       <h2>No schedule data.</h2>
       <p>
-        Please import a CSV of an existing schedule or start building the schedule from scratch by
+        Please import an existing schedule or start building the schedule from scratch by
         adding your first section.
       </p>
       <Box component="span" marginRight={1}>
@@ -18,7 +18,7 @@ export const NoCoursesHeader = () => {
       <Box component="span" marginRight={1}>
         <AddSectionButton isIcon={false} />
       </Box>
-      <ViewDemoButton />
+      {/* <ViewDemoButton /> */}
     </>
   );
 };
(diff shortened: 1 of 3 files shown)


rpruim:
I don't believe we support using the CSV from the new report at reports.calvin.edu.  The CSV format must come from the OLD form.  I don't think we need to bother mentioning how to get an existing schedule in CSV format.  We can just point people to the new workflow using the new report and Excel export.

Here is a possible rewrite:

To edit an existing schedule:

1. Obtain a previous schedule in one of three ways:
    a. Use a CSV file previously exported from an older version Course Schedulizer (before April 17, 2025), or
    b. Use an Excel file exported from a newer version of Course Schedulizer (after April 17, 2025),
    c. Export a schedule (in Excel format) using the “Schedulizer Course Sections” report from reports.calvin.edu.

2. Load the schedule file by clicking either:
        a. The “Import Schedule” button on the main startup page, or 
        b. The “Hamburger menu” (three lines in the upper left corner of the screen) and selecting “Import New Schedule”.

3. Use the standard schedule editing features to make the required changes.

rpruim:
Do we want to (a) make it clear that the current "formatting" section is about the old CSV files and (b) add a new section about the format required in the new Excel Format?

Do we want to include a warning disuading people from editing the Excel file out side of the Course Schedulizer?

kvlinden:
I think that we should just dump the "Formatting" section. Encouraging them to edit the excel or csv files directly is probably dangerous. Let me make some updates.

kvlinden:
So, there's a streamlined version of the help, without the field list.
better?

rpruim:
I think it looks good.  I like the warning about editing the file directly.

rpruim (approved):
I think the updated version looks good.

rpruim:
I've approved, but we could fold in another update.  On the initial splash page it says "Please import a CSV of an existing schedule or start building the schedule from scratch by adding your first section."  We should make that "Please import an existing schedule or ..."

kvlinden:
> I think it looks good. I like the warning about editing the file directly.

Copilot mostly wrote that :-)

kvlinden:
> I've approved, but we could fold in another update. On the initial splash page it says "Please import a CSV of an existing schedule or start building the schedule from scratch by adding your first section." We should make that "Please import an existing schedule or ..."

Updated...
And I commented out the demo button, which doesn't work at the moment for some reason...

PR #382, opened 2025-04-17
Merge current updates into production

kvlinden:
This will merge ~2 years of updates from develop into production. Updates include:

- Support for XLSX files, both import and export, based on requests from the registrar
- Newer versions of many of the packages
- Updated documentation

The merge will not include the new schedule-comparison feature (feature/compare), which will get merged into develop after this merge.


neh8 (approved):
It looks good!

PR #383, opened 2025-04-17
Resetting CI so that warnings are not treated as errors on pnpm build

kvlinden:
Turning off CI so that warnings don't create errors


neh8 (approved):
Looks good!

Issue #392, opened 2025-10-02
Check how Excel export handles a class with multiple times/rooms/etc.


rpruim:
Currently, the best way to handle such a section is to enter it twice, once for each time/room combo.

Things to check:

* [ ] export saves this correctly as one section with multiple meeting times/places (should be the same format as reports.calvin.edu exports; use a CS scheudule as an example)
* [ ] re-import maintains the look in Schedulizer that it had before export (currently broken)
* [ ] faculty load counts are correct.

rpruim:
see #393 as well.

kvlinden:
Note that the CSV export handles the multi-time/room courses properly. So that code could be used as a model for the XLSX code.

kvlinden:
Note also that I searched back in the dev and prod histories over a year can can't find a version that handles this properly with XLSX output.

PR #394, opened 2025-10-16
Package update

MinjiH67:
update react-stick, papaparse, fullcalendar/core, and fullcalendar/daygrid


kvlinden:
I don't see any new problems here. It still doesn't handle multiple time/room classes, but that's a known problem that we'll deal with soon (#392). 
There are some conflicts with pnpm-lock.yaml that need to be addressed, probably related to the version numbers.

PR #398, opened 2025-11-11
Add section fixed

MinjiH67:
When [addToSchedule] is called, it first creates a copy of the existing schedules array. The main change is that, if there are no schedules yet, the function now initializes the array with a new schedule based on either the current display schedule or a fresh empty schedule structure. 

By providing this default initialization, the system now has a valid target schedule to work with, even when no schedules have been imported.


kvlinden (approved):
LGTM
I testing the add section on a new schedule, and I testing loading an old schedule.
Both worked...

PR #399, opened 2025-11-13
add a test file for import schedule

MinjiH67:


MinjiH67:
Does this mean that I can't use Playwright with Netlify?

kvlinden (commented):
The error suggests that there are no meaningful changes to the running system - "Canceled build due to no content change". That would be consistent with the addition only a test file. And Netlify is still running. 

Do the tests run locally for you?

Also, what's the difference between the develop and the development branches. It looks like the former is the real branch and that the latter is mis-named. ?

PR #401, opened 2025-12-04
format for xlsx export is now fixed

MinjiH67:


kvlinden:
You'd better check the output on the "Registrar Schedule" tab. That needs to be updated too.

kvlinden (approved):
ok, that appears to be working!

PR #411, opened 2026-02-24
Feature/test minji

MinjiH67:
This is to test and compare the internal structure of the imported and exported files. However, it does not check individual values yet.


kvlinden:
Do you want to merge this now, or get the individual values tested first?
Also, was there some issue with the appropriate columns for the XLSX files that we needed to check with CIT? I can't remember what we said there.

rpruim:
I think the issue was that the report export doesn't include delivery mode (online, etc.).

kvlinden (approved):
LGTM

PR #414, opened 2026-04-07
Deprecated material-ui updated to muiv7

MinjiH67:
Package update branch

Opening the Schedulizer using different browsers before merging is needed.


kvlinden (approved):
LGTM

PR #416, opened 2026-04-19
Compare page new

Sylviali4:
Updated the Compare Feature with new Compare Setting page and Compare Result page, kept old compare feature files.


kvlinden:
The initial compare works as expected, but when I press "Modify Comparison Settings" and then immediately press "Show Comparison", without changing the settings, I get an "Invalid comparison settings", which goes away when I repeat those same steps again.
There must be a small bug in the saved settings here.

kvlinden:
Did this get fixed?

Sylviali4:
No, it’s not fixed, I tried some ways but still not working.

Lujia
________________________________

kvlinden:
Ok, we'll merge without a "fix" to this "feature/bug".

kvlinden (approved):
LGTM, modulo the bug listed in the PR.

